PR #43, opened 2022-09-02
[MI-2114] Modified the UI to support bulk subscriptons

manojmalik20:
Changed the header of the cards in the RHS to support bulk subscriptions along with the util function to get the header link
Created new type for SubscriptionType and did some refactoring of the types being used in the project
Changed the EditSubscriptionData along with the CreateSubscriptionData to include the subscription type
Created a new panel called SubscriptionTypePanel and integrated it in the Create/Edit subscription modal
Added the support for showing Record type in the Events panel


webapp/src/containers/addOrEditSubscriptions/subComponents/index.tsx
```diff
@@ -16,6 +16,7 @@ import usePluginApi from 'hooks/usePluginApi';
 import {refetch} from 'reducers/refetchSubscriptions';
 
 import ChannelPanel from './channelPanel';
+import SubscriptionTypePanel from './subscriptionTypePanel';
 import RecordTypePanel from './recordTypePanel';
 import EventsPanel from './eventsPanel';
 import SearchRecordsPanel from './searchRecordsPanel';
```
ayusht2810, line 22: is the import order correct here?

ayusht2810 (approved):

raghavaggarwal2308 (approved):

webapp/src/containers/addOrEditSubscriptions/subComponents/index.tsx
```diff
@@ -303,9 +322,9 @@ const AddOrEditSubscription = ({open, close, subscriptionData}: AddOrEditSubscri
             server_url: SiteURL ?? '',
             is_active: true,
             user_id: Cookies.get(Constants.MMUSERID) ?? '',
-            type: 'record',
-            record_type: recordType as string,
-            record_id: recordId as string,
+            type: subscriptionType as SubscriptionType,
+            record_type: recordType as RecordType,
+            record_id: recordId as string || '',
```
sooraj-shukla, line 327:
```suggestion
            record_id: recordId || '',
```

webapp/src/utils/index.ts
```diff
@@ -42,10 +42,14 @@ const debounce: (func: (args: Record<string, string>) => void, limit: number) =>
     };
 };
 
-const getSubscriptionNumberLink = (serviceNowBaseUrl: string, recordType: string, recordId: string): string => `${serviceNowBaseUrl}/nav_to.do?uri=${recordType}.do%3Fsys_id=${recordId}%26sysparm_stack=${recordType}_list.do%3Fsysparm_query=active=true`;
+const getSubscriptionHeaderLink = (serviceNowBaseUrl: string, subscriptionType: SubscriptionType, recordType: RecordType, recordId: string): string => (
+    subscriptionType === 'record' ?
```
sooraj-shukla, line 46: `Constants/Enum` for `subscriptionType`.

sooraj-shukla (approved):

webapp/src/containers/addOrEditSubscriptions/subComponents/index.tsx
manojmalik20: Fixed.

webapp/src/utils/index.ts
sooraj-shukla: I guess `enum` will be good.

anmolghosh (approved):

webapp/src/plugin_constants/index.ts
```diff
@@ -25,19 +25,30 @@ const ApiErrorIdSubscriptionsNotConfigured = 'subscriptions_not_configured';
...
-    state = 'state',
-    priority = 'priority',
-    commented = 'commented',
-    assignedTo = 'assigned_to',
-    assignmentGroup = 'assignment_group',
+    STATE = 'state',
+    PRIORITY = 'priority',
+    COMMENTED = 'commented',
+    ASSIGNED_TO = 'assigned_to',
+    ASSIGNMENT_GROUP = 'assignment_group',
```
avas27JTG, line 32: This applies only if we have used a mixture of camelCase or snake_case for request/response payload between plugin API and Webapp
I think we should use either only camelCase or snake_case for consistency
and if there is any inconsistency on the ServiceNow API i.e. it takes camel or snake for different API then we can handle this on plugin API

webapp/src/containers/addOrEditSubscriptions/subComponents/index.tsx
```diff
@@ -349,14 +368,26 @@ const AddOrEditSubscription = ({open, close, subscriptionData}: AddOrEditSubscri
...
+                />
                 <RecordTypePanel
                     className={`
                         ${recordTypePanelOpen && 'secondary-panel--slide-in'}
                         ${(searchRecordsPanelOpen || eventsPanelOpen) && 'secondary-panel--fade-out'}
                         ${(successPanelOpen || (apiResponseValid && apiError)) && 'secondary-panel--fade-out'}
                     `}
                     ref={recordTypePanelRef}
-                    onContinue={() => setSearchRecordsPanelOpen(true)}
+                    onContinue={() => (subscriptionType === SubscriptionType.RECORD ? setSearchRecordsPanelOpen(true) : setEventsPanelOpen(true))}
```
avas27JTG, line 390:
```
onContinue={() => subscriptionType === SubscriptionType.RECORD ? setSearchRecordsPanelOpen(true) : setEventsPanelOpen(true)}
```

webapp/src/plugin_constants/index.ts
manojmalik20: I just think keeping the casing of enum variables as all caps makes it look better because otherwise, there's doesn't seem to be a difference between the enum keys and the values.

webapp/src/containers/addOrEditSubscriptions/subComponents/index.tsx
manojmalik20: Removing the brackets here gives lint error.

manojmalik20: Removing the brackets here gives lint error.

PR #45, opened 2022-09-06
[MI-2118] Added the logic for returning sorted subscriptions from the…

manojmalik20:
… API

Reverted the order of bulk and record subscriptions in the List subscriptions slash command
Added an additional filter in the client function to get all subscriptions
Added the logic to segregate the bulk and record subscriptions in the API function


ayusht2810 (approved):

raghavaggarwal2308 (approved):

server/util.go
```diff
@@ -25,18 +25,18 @@ func ParseSubscriptionsToCommandResponse(subscriptions []*serializer.Subscriptio
...
-	}
-
 	if bulkSubscriptions.Len() > 0 {
-		sb.WriteString("\n#### Bulk subscriptions for this channel\n")
+		sb.WriteString("#### Bulk subscriptions for this channel\n")
 		sb.WriteString("| Subscription ID | Record Type | Events |\n| :----|:--------| :--------|")
 		sb.WriteString(bulkSubscriptions.String())
 	}
 
+	if recordSubscriptions.Len() > 0 {
```
avas27JTG, line 34: do we also need to handle a case for `len <= 0`?

manojmalik20: That is already handled.

avas27JTG (approved):

PR #49, opened 2022-09-14
[MI-2149] Add test cases for client

ayusht2810:


server/client_test.go
```diff
@@ -0,0 +1,387 @@
...
+	} {
+		t.Run(testCase.description, func(t *testing.T) {
+			monkey.PatchInstanceMethod(reflect.TypeOf(c), "CallJSON", func(_ *client, _, _ string, _, _ interface{}, _ url.Values) (_ []byte, _ int, _ error) {
+				return nil, testCase.statusCode, testCase.err
+			})
+
+			statusCode, err := c.ActivateSubscriptions()
+
+			if testCase.expectedError != nil {
+				assert.EqualError(t, testCase.expectedError, err.Error())
```
Kshitij-Katiyar, line 63: here why there is need to convert err to string and have expectedError as error?we can just have the opposite without converting anything. nad check this everywhere

ayusht2810: we need to convert `testCase.expectedError` to string then.

Kshitij-Katiyar: expectedError you are defining this you can just intitalise this as a string

ayusht2810: but shouldn't we define errors as `errors`?

Kshitij-Katiyar: yeah but that error you are creating with a string then converting to string again is a redundant

Kshitij-Katiyar (approved):

server/plugin/client_test.go
```diff
@@ -0,0 +1,387 @@
...
+			description:   "ActivateSubscriptions: user not authorized with error",
+			statusCode:    http.StatusForbidden,
+			err:           errors.New("mockError: User Not Authorized"),
+			expectedError: constants.APIErrorIDSubscriptionsNotAuthorized,
+		},
+		{
+			description:   "ActivateSubscriptions: user not authorized with status forbidden",
+			statusCode:    http.StatusForbidden,
+			err:           errors.New("mockError"),
+			expectedError: constants.APIErrorIDSubscriptionsNotAuthorized,
```
manojmalik20, line 40: I don't think there's a need for 2 test cases here. Both of these cover the same case.

server/plugin/client_test.go
```diff
@@ -0,0 +1,387 @@
...
+			statusCode:    http.StatusInternalServerError,
+			err:           errors.New("mockError: Invalid table"),
+			expectedError: constants.APIErrorIDSubscriptionsNotConfigured,
+		},
+		{
+			description:   "ActivateSubscriptions: failed to get subscription auth details",
+			statusCode:    http.StatusInternalServerError,
+			err:           errors.New("mockError"),
+			expectedError: "failed to get subscription auth details: mockError",
+		},
```
manojmalik20, line 53: All the cases are not covered here. For eg - when the CallJSON doesn't return any error for the first time but returns an error when it is called 2nd time in the function.

server/plugin/client_test.go
```diff
@@ -0,0 +1,387 @@
...
+			expectedErr: "failed to create subscription in ServiceNow: mockError",
+		},
+	} {
+		t.Run(testCase.description, func(t *testing.T) {
+			monkey.PatchInstanceMethod(reflect.TypeOf(c), "CallJSON", func(_ *client, _, _ string, _, _ interface{}, _ url.Values) (_ []byte, _ int, _ error) {
+				return nil, testCase.statusCode, testCase.err
+			})
+
+			statusCode, err := c.CreateSubscription(&serializer.SubscriptionPayload{})
+
```
manojmalik20, line 99:
```suggestion
```

server/plugin/client_test.go
```diff
@@ -0,0 +1,387 @@
...
+			expectedError: "failed to get subscription auth details: mockError",
+		},
+	} {
+		t.Run(testCase.description, func(t *testing.T) {
+			monkey.PatchInstanceMethod(reflect.TypeOf(c), "CallJSON", func(_ *client, _, _ string, _, _ interface{}, _ url.Values) (_ []byte, _ int, _ error) {
+				return nil, testCase.statusCode, testCase.err
+			})
+
+			statusCode, err := c.ActivateSubscriptions()
+
```
manojmalik20, line 61:
```suggestion
```

server/plugin/client_test.go
```diff
@@ -0,0 +1,387 @@
...
+			expectedErr: "failed to delete subscription from ServiceNow: mockError",
+		},
+	} {
+		t.Run(testCase.description, func(t *testing.T) {
+			monkey.PatchInstanceMethod(reflect.TypeOf(c), "CallJSON", func(_ *client, _, _ string, _, _ interface{}, _ url.Values) (_ []byte, _ int, _ error) {
+				return nil, testCase.statusCode, testCase.err
+			})
+
+			statusCode, err := c.DeleteSubscription("mockSubscriptionID")
+
```
manojmalik20, line 214:
```suggestion
```

server/plugin/client_test.go
```diff
@@ -0,0 +1,387 @@
...
+			expectedErr: "failed to update subscription from ServiceNow: mockError",
+		},
+	} {
+		t.Run(testCase.description, func(t *testing.T) {
+			monkey.PatchInstanceMethod(reflect.TypeOf(c), "CallJSON", func(_ *client, _, _ string, _, _ interface{}, _ url.Values) (_ []byte, _ int, _ error) {
+				return nil, testCase.statusCode, testCase.err
+			})
+
+			statusCode, err := c.EditSubscription("mockSubscriptionID", &serializer.SubscriptionPayload{})
+
```
manojmalik20, line 252:
```suggestion
```

server/plugin/client_test.go
```diff
@@ -0,0 +1,387 @@
...
+			expectedErr: "mockError",
+		},
+	} {
+		t.Run(testCase.description, func(t *testing.T) {
+			monkey.PatchInstanceMethod(reflect.TypeOf(c), "CallJSON", func(_ *client, _, _ string, _, _ interface{}, _ url.Values) (_ []byte, _ int, _ error) {
+				return nil, testCase.statusCode, testCase.err
+			})
+
+			_, statusCode, err := c.GetRecordFromServiceNow("mockTable", "mockSysID")
+
```
manojmalik20, line 377:
```suggestion
```

server/plugin/client_test.go
ayusht2810: we already have mocked the CallJSON method once. If we return an error, it will be from the first call of CallJSON itself.

manojmalik20 (approved):

PR #54, opened 2022-09-16
Base setup for unit tests and some tests for api.go

manojmalik20:


server/testutils/data.go
```diff
@@ -0,0 +1,104 @@
...
+
+func GetChannels(count int, channelType string) []*model.Channel {
+	if count == 0 {
+		return nil
+	}
+
+	if channelType == "" {
+		channelType = model.CHANNEL_OPEN
+	}
+	channels := make([]*model.Channel, count)
```
raghavaggarwal2308, line 36:
```suggestion
	
	channels := make([]*model.Channel, count)
```

server/plugin/api_test.go
```diff
@@ -0,0 +1,347 @@
...
+			r := httptest.NewRequest(requestMethod, strings.Replace(requestURL, "{team_id:[A-Za-z0-9]+}", test.TeamID, 1), nil)
+			r.Header.Add(constants.HeaderMattermostUserID, testutils.GetID())
+			p.ServeHTTP(nil, w, r)
+
+			result := w.Result()
+			require.NotNil(t, result)
+			defer result.Body.Close()
+
+			assert.Equal(test.ExpectedStatusCode, result.StatusCode)
+			if test.ExpectedCount != -1 {
```
raghavaggarwal2308, line 166:
```suggestion
			
			if test.ExpectedCount != -1 {
```

raghavaggarwal2308 (approved):

server/plugin/api_test.go
```diff
@@ -0,0 +1,347 @@
...
+	"golang.org/x/oauth2"
+)
+
+func setupTestPlugin(api *plugintest.API, store *mock_plugin.Store) *Plugin {
+	p := &Plugin{}
+	p.setConfiguration(&configuration{
+		WebhookSecret: testutils.GetSecret(),
+	})
+
+	path, _ := filepath.Abs("../..")
```
Kshitij-Katiyar, line 33: I think here this function could return an error so we should handle this

server/testutils/data.go
```diff
@@ -0,0 +1,104 @@
...
+		records[i] = GetServiceNowPartialRecord()
+	}
+
+	return records
+}
+
+func GetSerializerUser() *serializer.User {
+	return &serializer.User{
+		MattermostUserID: GetID(),
+		OAuth2Token:      "adfafdafadsfas",
```
Kshitij-Katiyar, line 98: if we are giving a random value then it could be related to actual variable like :- 
```suggestion
		OAuth2Token:      "test-oauthtoken",
```

server/testutils/data.go
```diff
@@ -0,0 +1,104 @@
...
+	"net/http"
+
+	"github.com/Brightscout/mattermost-plugin-servicenow/server/constants"
+	"github.com/Brightscout/mattermost-plugin-servicenow/server/serializer"
+	"github.com/mattermost/mattermost-server/v5/api4"
+	"github.com/mattermost/mattermost-server/v5/model"
+)
+
+func GetSecret() string {
+	return "1234567890abcdefghij"
```
Kshitij-Katiyar, line 14:
```suggestion
	return "test-secret"
```

server/plugin/api_test.go
manojmalik20: 1. I don't think it will return an error in any case.
2. If we were to handle the error here, how should we handle it?

Kshitij-Katiyar (approved):

server/plugin/api_test.go
```diff
@@ -0,0 +1,349 @@
...
+		"invalid search term": {
+			RecordType: constants.SubscriptionRecordTypeIncident,
+			SearchTerm: "sdf",
+			SetupAPI: func(api *plugintest.API) *plugintest.API {
+				return api
+			},
+			SetupClient:          func(p *Plugin) {},
+			ExpectedStatusCode:   http.StatusBadRequest,
+			ExpectedCount:        -1,
+			ExpectedErrorMessage: "The search term must be at least 4 characters long.",
```
avas27JTG, line 237: we should change this test flow as I think the count of characters is modified now

avas27JTG (approved):

PR #66, opened 2022-10-03
[MI-2213]: Update ChanelPanel, fix lint issues

sooraj-shukla:


webapp/src/containers/addOrEditSubscriptions/subComponents/channelPanel.tsx
```diff
@@ -107,6 +110,13 @@ const ChannelPanel = forwardRef<HTMLDivElement, ChannelPanelProps>(({
         }
     }, [channel]);
 
+    // Provide the default value when subscription is being edited
```
ayusht2810, line 113: I don't think we now need lines 80-82.
`if (channel) {
                setChannelAutoSuggestValue(getRequiredChannelName(channel, channelListState.data));
            }`
            
 please remove this if we don't need these

webapp/src/containers/addOrEditSubscriptions/subComponents/channelPanel.tsx
```diff
@@ -107,6 +110,13 @@ const ChannelPanel = forwardRef<HTMLDivElement, ChannelPanelProps>(({
...
+    useEffect(() => {
+        if (editing && getChannelState().isSuccess) {
+            setAutoSuggestDefaultValue(mapChannelsToSuggestions(getChannelState()?.data?.filter((ch) => ch.id === channel) as unknown as ChannelData[])?.[0]);
```
manojmalik20, line 116: Use `find` instead of `filter` here.

manojmalik20 (approved):

webapp/src/containers/addOrEditSubscriptions/subComponents/channelPanel.tsx
sooraj-shukla: I intentionally used `filter` here as if I used `find` we'll have to convert that one item(which will be of type `undefined | ChannelData`) into an array (which should be of type `ChannelData[ ]`) again to pass that as the argument to mapChannelsToSuggestions() function. We can ignore these steps just by using `filter` instead of `find`.

manojmalik20 (approved):

PR #67, opened 2022-10-06
[MI-2210] Added the API for getting comments on a record

manojmalik20:
Added serializer for ServiceNowComment
Created client function for fetching comments on a record Created constants and a paginated API for getting comments on a record Updated the mock client generated by mockery


server/plugin/api.go
```diff
@@ -567,6 +568,44 @@ func (p *Plugin) shareRecordInChannel(w http.ResponseWriter, r *http.Request) {
 	returnStatusOK(w)
 }
 
+func (p *Plugin) getCommentsForRecord(w http.ResponseWriter, r *http.Request) {
+	pathParams := mux.Vars(r)
+	recordType := pathParams[constants.PathParamRecordType]
+	if !constants.ValidRecordTypesForSearching[recordType] {
+		p.API.LogError("Invalid record type while trying to get record", "Record type", recordType)
```
Kshitij-Katiyar, line 575:
```suggestion
		p.API.LogError("Invalid record type while trying to get the record", "Record type", recordType)
```

server/plugin/api.go
```diff
@@ -567,6 +568,44 @@ func (p *Plugin) shareRecordInChannel(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	page, perPage := GetPageAndPerPage(r)
+	commentsArray := ProcessComments(comments, page, perPage)
+
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(statusCode)
+	result, err := json.Marshal(commentsArray)
+	if err != nil {
+		p.API.LogDebug("Error while marshaling the response", "Error", err.Error())
```
Kshitij-Katiyar, line 596:
```suggestion
		p.API.LogDebug("Error while marshalling the response", "Error", err.Error())
```

server/plugin/client.go
```diff
@@ -187,11 +188,27 @@ func (c *client) GetRecordFromServiceNow(tableName, sysID string) (*serializer.S
...
 		return nil, statusCode, err
 	}
 
 	return record.Result, statusCode, nil
 }
+
+func (c *client) GetAllComments(recordType, recordID string) (string, int, error) {
+	queryParams := url.Values{
+		constants.SysQueryParamDisplayValue: {"true"},
+		constants.SysQueryParamFields:       {"comments_and_work_notes"},
```
Kshitij-Katiyar, line 203: we should create a constant for this "comments_and_work_notes"

server/plugin/api.go
manojmalik20: Actually, the correct spelling is the one containing single `l`.

Kshitij-Katiyar: i researched a bit its usa vs uk english so i guess both are fine

Kshitij-Katiyar (approved):

server/plugin/api.go
```diff
@@ -567,6 +568,44 @@ func (p *Plugin) shareRecordInChannel(w http.ResponseWriter, r *http.Request) {
...
+		p.API.LogError("Invalid record type while trying to get the record", "Record type", recordType)
+		p.handleAPIError(w, &serializer.APIErrorResponse{StatusCode: http.StatusBadRequest, Message: constants.ErrorInvalidRecordType})
+		return
+	}
+
+	recordID := pathParams[constants.PathParamRecordID]
+	client := p.GetClientFromRequest(r)
+	comments, statusCode, err := client.GetAllComments(recordType, recordID)
+	if err != nil {
+		p.API.LogError("Error in getting all comments", "Record ID", recordID, "Error", err.Error())
```
avas27JTG, line 584: can create constants for all messages like `Error in getting all comments` as there are chances they are used at many places or in test cases

avas27JTG: If we create constants for some messages then I think it would be good to create constants and keep all messages together for code consistency, we can add TODO to move all messages in a separate file under the same `constant` package as we did in Azure DevOps, however on Azure also some messages are inline which need to be moved.

avas27JTG (approved):

PR #70, opened 2022-10-07
[MI-2227] Add feature to filter subscriptions using slash command

ayusht2810:


server/plugin/command.go
```diff
@@ -262,6 +284,22 @@ func (p *Plugin) handleListSubscriptions(_ *plugin.Context, args *model.CommandA
 
 		wg := sync.WaitGroup{}
 		for _, subscription := range subscriptions {
+			user, err := p.API.GetUser(subscription.UserID)
+			if err != nil {
+				p.API.LogError("User not found with user ID %d", subscription.UserID)
```
raghavaggarwal2308, line 289:
```suggestion
				p.API.LogError("User not found with userID %d", subscription.UserID)
```

server/plugin/command.go
```diff
@@ -262,6 +284,22 @@ func (p *Plugin) handleListSubscriptions(_ *plugin.Context, args *model.CommandA
...
+				subscription.UserName = "N/A"
+			} else {
+				subscription.UserName = user.Username
+			}
+
+			channel, err := p.API.GetChannel(subscription.ChannelID)
+			if err != nil {
+				p.API.LogError("Channel not found with channel ID %d", subscription.ChannelID)
```
raghavaggarwal2308, line 297:
```suggestion
				p.API.LogError("Channel not found with channelID %d", subscription.ChannelID)
```

raghavaggarwal2308 (approved):

server/plugin/command.go
```diff
@@ -246,9 +246,31 @@ func (p *Plugin) handleSubscribe(_ *plugin.Context, args *model.CommandArgs, par
 	return ""
 }
 
-func (p *Plugin) handleListSubscriptions(_ *plugin.Context, args *model.CommandArgs, _ []string, client Client) string {
+func (p *Plugin) handleListSubscriptions(_ *plugin.Context, args *model.CommandArgs, params []string, client Client) string {
+	if len(params) < 1 {
+		return "Invalid number of params for this command."
```
manojmalik20, line 251: Instead of doing this, shouldn't we have a default behaviour for the command in case no param is given? :thinking:

server/plugin/command.go
```diff
@@ -246,9 +246,31 @@ func (p *Plugin) handleSubscribe(_ *plugin.Context, args *model.CommandArgs, par
...
+	}
+
+	if params[0] != constants.FilterCreatedByMe && params[0] != constants.FilterCreatedByAnyone {
+		return fmt.Sprintf("Unknown filter %s", params[0])
+	}
+
+	userID := ""
+	channelID := args.ChannelId
+
```
manojmalik20, line 260:
```suggestion
```

server/plugin/command.go
```diff
@@ -262,6 +284,22 @@ func (p *Plugin) handleListSubscriptions(_ *plugin.Context, args *model.CommandA
 
 		wg := sync.WaitGroup{}
 		for _, subscription := range subscriptions {
+			user, err := p.API.GetUser(subscription.UserID)
+			if err != nil {
+				p.API.LogError("User not found with userID %d", subscription.UserID)
```
manojmalik20, line 289:
```suggestion
				p.API.LogError("Error in getting user", "UserID", subscription.UserID)
```

server/plugin/command.go
```diff
@@ -262,6 +284,22 @@ func (p *Plugin) handleListSubscriptions(_ *plugin.Context, args *model.CommandA
...
+				subscription.UserName = "N/A"
+			} else {
+				subscription.UserName = user.Username
+			}
+
+			channel, err := p.API.GetChannel(subscription.ChannelID)
+			if err != nil {
+				p.API.LogError("Channel not found with channelID %d", subscription.ChannelID)
```
manojmalik20, line 297: Same.

server/plugin/command.go
```diff
@@ -262,6 +284,22 @@ func (p *Plugin) handleListSubscriptions(_ *plugin.Context, args *model.CommandA
 
 		wg := sync.WaitGroup{}
 		for _, subscription := range subscriptions {
+			user, err := p.API.GetUser(subscription.UserID)
```
manojmalik20, line 287: You can do this `GetUser` and `GetChannel` in a separate goroutine to make this faster just like it is being done at line 307.

server/serializer/subscription.go
```diff
@@ -25,7 +25,9 @@ type SubscriptionPayload struct {
 type SubscriptionResponse struct {
 	SysID              string `json:"sys_id"`
 	UserID             string `json:"user_id"`
+	UserName           string `json:"userName"`
```
manojmalik20, line 28:
```suggestion
	UserName           string `json:"-"`
```
Same for the channelName.

server/serializer/subscription.go
```diff
@@ -36,7 +38,7 @@ type SubscriptionResponse struct {
 	ShortDescription   string `json:"short_description"`
 }
 
-func (s *SubscriptionResponse) GetFormattedSubscription() string {
+func (s *SubscriptionResponse) GetFormattedSubscription(user, channel string) string {
```
manojmalik20, line 41: What's the need of taking params when these values are present inside the `SubscriptionResponse`?

server/plugin/command.go
ayusht2810: updated this. if there are no params, we will be showing the user's current channel subscriptions

server/plugin/command.go
```diff
@@ -284,22 +281,27 @@ func (p *Plugin) handleListSubscriptions(_ *plugin.Context, args *model.CommandA
...
+				channel, err := p.API.GetChannel(subscription.ChannelID)
+				if err != nil {
+					p.API.LogError("Error in getting channel", "ChannelID", subscription.ChannelID)
+					subscription.ChannelName = "N/A"
+				} else {
+					subscription.ChannelName = channel.DisplayName
+				}
+			}()
 
+			wg.Wait()
```
manojmalik20, line 304:
```suggestion
```

ayusht2810: Tried removing this already, but was getting fields empty for some subscriptions

manojmalik20: Then you should debug that and find the cause for that. If we add the `wait` statement, we're not really optimizing the speed.

manojmalik20 (approved):

PR #71, opened 2022-10-07
[MI-2229] Added an API to add a comment on a record

manojmalik20:
Created the client function for adding a comment
Created the serializer for comment payload and the route for the API


ayusht2810 (approved):

raghavaggarwal2308 (approved):

server/plugin/api.go
```diff
@@ -606,6 +607,34 @@ func (p *Plugin) getCommentsForRecord(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func (p *Plugin) addCommentsOnRecord(w http.ResponseWriter, r *http.Request) {
+	pathParams := mux.Vars(r)
+	recordType := pathParams[constants.PathParamRecordType]
+	if !constants.ValidRecordTypesForSearching[recordType] {
+		p.API.LogError("Invalid record type while trying to get the record", "Record type", recordType)
```
avas27JTG, line 615:
```suggestion
		p.API.LogError("invalid record type while trying to get the record", "Record type", recordType)
```

server/plugin/api.go
```diff
@@ -606,6 +607,34 @@ func (p *Plugin) getCommentsForRecord(w http.ResponseWriter, r *http.Request) {
...
+		p.handleAPIError(w, &serializer.APIErrorResponse{StatusCode: http.StatusBadRequest, Message: constants.ErrorInvalidRecordType})
+		return
+	}
+
+	payload, err := serializer.ServiceNowCommentPayloadFromJSON(r.Body)
+	if err != nil {
+		p.API.LogError("Error in unmarshalling the request body", "Error", err.Error())
```
avas27JTG, line 622:
```suggestion
		p.API.LogError("error in unmarshalling the request body", "Error", err.Error())
```
I think messages that we are logging on the server start with a small letter and which are passed to UI only start with a capital

manojmalik20: No @avas27JTG , the logging messages also start with a capital letter. You can check in the whole repo.

avas27JTG: Okay, then I think I wasn't aware of it, and I was thinking we add small on logs and capital in UI
can you please post the rule that we follow for small/capital messages on MM so that every one of us can read that when needed?

avas27JTG (approved):

server/plugin/api.go
manojmalik20: Sure, I'll post that.

PR #77, opened 2022-10-12
[MI-2246] Added the API to update the state of a record

manojmalik20:
Updated the constant for the record types that support state updation Changed the validation of record type in the comments API Created the client function and the API for updating the state Created the serializer for state payload and added validation on it

### DIFF
--- a/server/constants/routes.go
+++ b/server/constants/routes.go
@@ -18,6 +18,7 @@ const (
 	PathShareRecord            = "/share/{channel_id:[A-Za-z0-9]+}"
 	PathCommentsForRecord      = "/comments/{record_type}/{record_id:" + ServiceNowSysIDRegex + "}"
 	PathGetStatesForRecordType = "/states/{record_type}"
+	PathUpdateStateOfRecord    = "/states/{record_type}/{record_id:" + ServiceNowSysIDRegex + "}"
 
 	// ServiceNow API paths
 	PathActivateSubscriptions    = "api/now/table/" + ServiceNowForMattermostNotificationsAppID + "_servicenow_for_mattermost_notifications_auth"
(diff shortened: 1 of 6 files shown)


ayusht2810 (approved):

Kshitij-Katiyar (approved):

server/constants/routes.go
```diff
@@ -18,6 +18,7 @@ const (
 	PathShareRecord            = "/share/{channel_id:[A-Za-z0-9]+}"
 	PathCommentsForRecord      = "/comments/{record_type}/{record_id:" + ServiceNowSysIDRegex + "}"
 	PathGetStatesForRecordType = "/states/{record_type}"
+	PathUpdateStateOfRecord    = "/states/{record_type}/{record_id:" + ServiceNowSysIDRegex + "}"
```
avas27JTG, line 21:
```suggestion
	PathUpdateStateOfRecord    = fmt.Sprintf( "/states/{record_type}/{record_id:%s}", ServiceNowSysIDRegex)
```

can we do this?

server/plugin/api.go
```diff
@@ -612,8 +613,8 @@ func (p *Plugin) getCommentsForRecord(w http.ResponseWriter, r *http.Request) {
 func (p *Plugin) addCommentsOnRecord(w http.ResponseWriter, r *http.Request) {
 	pathParams := mux.Vars(r)
 	recordType := pathParams[constants.PathParamRecordType]
-	if !constants.ValidRecordTypesForSearching[recordType] {
-		p.API.LogError("Invalid record type while trying to get the record", "Record type", recordType)
+	if !constants.RecordTypesSupportingComments[recordType] {
```
avas27JTG, line 616:
```suggestion
	if !constants.RecordTypeSupportingComments[recordType] {
```

server/plugin/api.go
```diff
@@ -640,12 +641,16 @@ func (p *Plugin) addCommentsOnRecord(w http.ResponseWriter, r *http.Request) {
 func (p *Plugin) getStatesForRecordType(w http.ResponseWriter, r *http.Request) {
 	pathParams := mux.Vars(r)
 	recordType := pathParams[constants.PathParamRecordType]
-	if !constants.ValidRecordTypesForSearching[recordType] {
-		p.API.LogError("Invalid record type while trying to get the record", "Record type", recordType)
+	if !constants.RecordTypesSupportingStateUpdation[recordType] {
```
avas27JTG, line 644:
```suggestion
	if !constants.RecordTypeSupportingStateUpdation[recordType] {
```
should not we remove 's' from the 'types' as at a time it's targeting/working for only 1 type?

server/constants/routes.go
manojmalik20: No, we can't use `Sprintf` in constants.

server/plugin/api.go
manojmalik20: That variable is a map that contains all the record types that support state updation. I don't think removing the 's' will be a better name for it.

PR #78, opened 2022-10-12
[MI-2249] Added a slash command for opening Search modal and refactored the slash commands

manojmalik20:
Created constants for all the slash commands and used them everywhere
Created slash command for search and added validation for connect in that
Created websocket event for opening the search and share modal and published it on slash command
Added websocket handler on the webapp to open the search and share modal


ayusht2810 (approved):

raghavaggarwal2308 (approved):

server/plugin/command.go
```diff
@@ -246,6 +247,16 @@ func (p *Plugin) handleSubscribe(_ *plugin.Context, args *model.CommandArgs, par
...
 }
 
+func (p *Plugin) handleSearchAndShare(_ *plugin.Context, args *model.CommandArgs, params []string, client Client) string {
+	p.API.PublishWebSocketEvent(
+		constants.WSEventOpenSearchAndShareRecordsModal,
+		nil,
+		&model.WebsocketBroadcast{UserId: args.UserId},
+	)
+
+	return ""
```
avas27JTG, line 257: why we need `return` here, can't we modify the func to remove return?

avas27JTG (approved):

PR #79, opened 2022-10-13
[MI-2251] Added the check in the GetStates API if the latest update set has been uploaded or not

manojmalik20:
Created constants for the API error ID and message if the GetStates scripted API is not found on ServiceNow
Modified the client function and the API function according to the error
Removed a TODO comment from the ServiceNowEvent serializer and added the latest update set

### DIFF
--- a/server/plugin/api.go
+++ b/server/plugin/api.go
@@ -649,6 +649,11 @@ func (p *Plugin) getStatesForRecordType(w http.ResponseWriter, r *http.Request) 
 	client := p.GetClientFromRequest(r)
 	states, statusCode, err := client.GetStatesFromServiceNow(recordType)
 	if err != nil {
+		if strings.EqualFold(err.Error(), constants.APIErrorIDLatestUpdateSetNotUploaded) {
+			p.handleAPIError(w, &serializer.APIErrorResponse{ID: constants.APIErrorIDLatestUpdateSetNotUploaded, StatusCode: statusCode, Message: constants.APIErrorLatestUpdateSetNotUploaded})
+			return
+		}
+
 		p.API.LogError("Error in getting the states", "Record Type", recordType, "Error", err.Error())
 		p.handleAPIError(w, &serializer.APIErrorResponse{StatusCode: statusCode, Message: fmt.Sprintf("Error in getting the states. Error: %s", err.Error())})
 		return
--- a/server/plugin/client.go
+++ b/server/plugin/client.go
@@ -227,6 +227,10 @@ func (c *client) GetStatesFromServiceNow(recordType string) ([]*serializer.Servi
 	url := strings.Replace(constants.PathGetStatesFromServiceNow, "{record_type}", recordType, 1)
 	_, statusCode, err := c.CallJSON(http.MethodGet, url, nil, states, nil)
 	if err != nil {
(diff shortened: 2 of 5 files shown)


server/plugin/api.go
```diff
@@ -649,6 +649,11 @@ func (p *Plugin) getStatesForRecordType(w http.ResponseWriter, r *http.Request)
 	client := p.GetClientFromRequest(r)
 	states, statusCode, err := client.GetStatesFromServiceNow(recordType)
 	if err != nil {
+		if strings.EqualFold(err.Error(), constants.APIErrorIDLatestUpdateSetNotUploaded) {
+			p.handleAPIError(w, &serializer.APIErrorResponse{ID: constants.APIErrorIDLatestUpdateSetNotUploaded, StatusCode: statusCode, Message: constants.APIErrorLatestUpdateSetNotUploaded})
```
ayusht2810, line 653: what is the use of `ID` here?

ayusht2810 (approved):

server/plugin/client.go
```diff
@@ -227,6 +227,10 @@ func (c *client) GetStatesFromServiceNow(recordType string) ([]*serializer.Servi
 	url := strings.Replace(constants.PathGetStatesFromServiceNow, "{record_type}", recordType, 1)
 	_, statusCode, err := c.CallJSON(http.MethodGet, url, nil, states, nil)
 	if err != nil {
+		if statusCode == http.StatusBadRequest && strings.Contains(err.Error(), "Requested URI does not represent any resource") {
```
raghavaggarwal2308, line 230:
```suggestion
		if statusCode == http.StatusBadRequest && strings.Contains(err.Error(), "Requested URI does not represent any resource") {
```
It should be `URI` or `URL`?

raghavaggarwal2308 (approved):

server/plugin/api.go
manojmalik20: It will used on the webapp side to distinguish the error and show the error state accordingly. For example, see this - https://github.com/Brightscout/mattermost-plugin-servicenow/blob/ae96b8fd1ae5a1da71575742894cd801fa55d3fa/webapp/src/containers/Rhs/index.tsx#L142

server/plugin/client.go
manojmalik20: `URI`. We get this response from ServiceNow.

server/serializer/servicenow_event.go
```diff
@@ -24,8 +24,7 @@ type ServiceNowEvent struct {
 	Priority         string `json:"priority"`
 	AssignedTo       string `json:"assigned_to"`
 	AssignmentGroup  string `json:"assignment_group"`
-	// TODO: Remove the nolint comment by changing spelling of occurred in the update set XML
-	EventOccured string `json:"event_occured"` // nolint
+	EventOccurred    string `json:"event_occurred"`
```
avas27JTG, line 27: how this is fixed on SN?

avas27JTG (approved):

server/serializer/servicenow_event.go
manojmalik20: This is fixed by changing the variable name in ServiceNow script and that change is present in the update set.

PR #80, opened 2022-10-13
[MI-2253] Added the logic to open Comment modal from the slack attachment post

manojmalik20:
Created an API to publish the websocket event for opening the comment modal
Added the comments and state buttons on the notification post also
Added the logic to register websocket event handler on the webapp and open the comment modal with the payload
Modified the reducer for comment modal a little and created a type


raghavaggarwal2308 (approved):

server/serializer/servicenow_event.go
```diff
@@ -50,6 +50,28 @@ func (se *ServiceNowEvent) CreateNotificationPost(botID, serviceNowURL string) *
...
+					constants.ContextNameRecordType: se.RecordType,
+					constants.ContextNameRecordID:   se.RecordID,
+				},
+			},
+		})
+	}
+
+	if constants.RecordTypesSupportingStateUpdation[se.RecordType] {
+		actions = append(actions, &model.PostAction{
+			Type: "button",
```
Kshitij-Katiyar, line 70: should be create a constant for button because it is a type for post action and we may need to use it in future

Kshitij-Katiyar (approved):

webapp/src/websocket/index.ts
```diff
@@ -68,3 +67,15 @@ export function handleOpenShareRecordModal(store: Store<GlobalState, Action<Reco
         store.dispatch(showShareRecordModal() as Action);
     };
 }
+
+export function handleOpenCommentModal(store: Store<GlobalState, Action<Record<string, unknown>>>) {
+    return (msg: WebsocketEventParams) => {
+        // Fix the type of state below by importing the GlobalState from mattermost-webapp
```
avas27JTG, line 73:
```suggestion
        // TODO: Fix the type of state below by importing the GlobalState from mattermost-webapp
```

avas27JTG (approved):

PR #81, opened 2022-10-14
[MI-2255] Add feature to filter subscriptions using rhs

ayusht2810:


webapp/src/components/Buttons/button/index.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+
+type PrimaryButtonProps = {
+    text: string
+    iconName?: string
+    extraClass?: string
+    onClick?: () => void
```
sooraj-shukla, line 7: `event` should be passed as the argument.

webapp/src/components/Buttons/iconButton/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+}
+
+const IconButton = ({tooltipText, iconClassName, children, extraClass = '', onClick}: IconButtonProps) => (
+    <Tooltip tooltipContent={tooltipText}>
+        <Button
+            variant='outline-danger'
+            className={`plugin-btn button-wrapper btn-icon ${extraClass}`}
+            onClick={onClick}
+            aria-label={tooltipText}
+            role='button'
```
sooraj-shukla, line 25: Is this really required to add `role` for button element in html5? 
Have a look at this - https://stackoverflow.com/questions/26934719/do-i-need-role-button-on-a-button

webapp/src/styles/base/_common.scss
```diff
@@ -1,3 +1,3 @@
 .rhs-state-icon {
-  opacity: 0.6;
+  opacity: 1;
```
sooraj-shukla, line 2: Do we really need to add `opacity: 1`? It's the default one. Are we overriding the style here?

sooraj-shukla (changes requested):

webapp/src/styles/base/_common.scss
ayusht2810: changed this as the bell icon was coming on the subscription popover

sooraj-shukla: No, by default the opacity is always 1. So, I was saying do we really need to include this now, can't we just remove the class?

ayusht2810: yes we are overriding the style here

sooraj-shukla (approved):

webapp/src/hooks/useClickOutside.ts
```diff
@@ -0,0 +1,22 @@
+import {useEffect} from 'react';
+
+// Hook that detects clicks outside of the passed ref
+function useOutsideClick(ref: React.RefObject<HTMLInputElement>, handleOnClickOutside: () => void) {
+    useEffect(() => {
+        // Check if clicked on outside of element
```
raghavaggarwal2308, line 6:
```suggestion
        // Check if clicked outside of the element
```

raghavaggarwal2308 (approved):

webapp/src/plugin_constants/index.ts
```diff
@@ -147,6 +147,28 @@ const RecordDataLabelConfig: RecordDataLabelConfigType[] = [
     },
 ];
 
+export const SubscriptionFilters = {
```
manojmalik20, line 150: I think this should be an enum.

webapp/src/components/Buttons/button/index.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+
+type PrimaryButtonProps = {
```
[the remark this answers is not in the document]
ayusht2810: removed this component now

manojmalik20 (approved):

PR #86, opened 2022-10-20
[MI-2265] Integrated the Update state modal

manojmalik20:
Added the logic to open the update state modal on the click of button in the post
Created services, constants and types needed for the API calls to get and update state
Added the logic to get the states from API and call the update state API in the modal
Added some additional styling to fix the UI of the modal
Added the logic to open the modal with some props like record type and record ID


server/plugin/api.go
```diff
@@ -616,6 +617,11 @@ func (p *Plugin) updateStateOfRecord(w http.ResponseWriter, r *http.Request) {
 	client := p.GetClientFromRequest(r)
 	statusCode, err := client.UpdateStateOfRecordInServiceNow(recordType, recordID, payload)
 	if err != nil {
+		if statusCode == http.StatusNotFound && strings.Contains(err.Error(), "ACL restricts the record retrieval") {
```
Kshitij-Katiyar, line 620: do we need a constant for this

webapp/src/containers/updateState/styles.scss
```diff
@@ -0,0 +1,7 @@
+.update-state-modal {
+    .wizard__secondary-panel {
+        position: static;
+        transform: translateX(0);
+        padding: 0;
+    }
+}
```
Kshitij-Katiyar, line 7: add EOF

webapp/src/containers/updateState/index.tsx
```diff
@@ -1,84 +1,142 @@
...
+
+    useEffect(() => {
+        const {data, open} = pluginState.openUpdateStateModalReducer;
+        if (!(open && data?.recordType && data?.recordId)) {
+            return;
+        }
+
+        const params: GetStatesParams = {recordType: data.recordType};
+        setGetStatesParams(params);
+        makeApiRequest(Constants.pluginApiServiceConfigs.getStates.apiServiceName, params);
```
ayusht2810, line 60:
```suggestion
        if (open && data?.recordType && data?.recordId) {
        const params: GetStatesParams = {recordType: data.recordType};
        setGetStatesParams(params);
        makeApiRequest(Constants.pluginApiServiceConfigs.getStates.apiServiceName, params);
        }
```

webapp/src/containers/updateState/index.tsx
```diff
@@ -1,84 +1,142 @@
...
+            setShowResultPanel(true);
+        }
+
+        if (isSuccess) {
+            setApiError('');
+        }
+    }, [getStateForGetStatesAPI().isError, getStateForGetStatesAPI().isSuccess]);
+
+    const {isLoading: statesLoading, data: stateOptions} = getStateForGetStatesAPI();
+    const {isLoading: stateUpdating} = getStateForUpdateStateAPI();
```
ayusht2810, line 96:
```suggestion
    const {isLoading: statesUpdating} = getStateForUpdateStateAPI();
```

webapp/src/websocket/index.ts
```diff
@@ -82,7 +82,13 @@ export function handleOpenCommentModal(store: Store<GlobalState, Action<Record<s
 }
 
 export function handleOpenUpdateStateModal(store: Store<GlobalState, Action<Record<string, unknown>>>) {
-    return (_: WebsocketEventParams) => {
-        store.dispatch(showUpdateStateModal() as Action);
+    return (msg: WebsocketEventParams) => {
+        // Fix the type of state below by importing the GlobalState from mattermost-webapp
```
ayusht2810, line 86: do we need this comment?

server/plugin/api.go
manojmalik20: It will be created while writing unit tests for this function.

webapp/src/containers/updateState/index.tsx
manojmalik20: We are using singular `state` for update state everywhere.

ayusht2810 (approved):

Kshitij-Katiyar (approved):

webapp/src/containers/updateState/index.tsx
```diff
@@ -1,84 +1,140 @@
...
+        if (isError) {
+            setApiError(error?.message ?? '');
+            setShowResultPanel(true);
+        }
+
+        if (isSuccess) {
+            setApiError('');
+            setShowResultPanel(true);
+        }
+    }, [getStateForUpdateStateAPI().isError, getStateForUpdateStateAPI().isSuccess]);
```
avas27JTG, line 79: you can use `useApiRequestCompletionState` implemented in AZD https://github.com/Brightscout/mattermost-plugin-azure-devops/blob/master/webapp/src/hooks/useApiRequestCompletionState.ts

webapp/src/containers/updateState/index.tsx
```diff
@@ -1,84 +1,140 @@
...
+        const {isError, isSuccess, error} = getStateForGetStatesAPI();
+        if (isError) {
+            setApiError(error?.message ?? '');
+            setShowResultPanel(true);
+        }
+
+        if (isSuccess) {
+            setApiError('');
+        }
+    }, [getStateForGetStatesAPI().isError, getStateForGetStatesAPI().isSuccess]);
```
avas27JTG, line 91: a lot of code can be reduced using https://github.com/Brightscout/mattermost-plugin-azure-devops/blob/master/webapp/src/hooks/useApiRequestCompletionState.ts

if it's not possible now due to the timeline can add `TODO` to revisit this

webapp/src/containers/updateState/index.tsx
```diff
@@ -1,84 +1,140 @@
...
+                {showLoader && <CircularLoader/>}
+                {showResultPanel ? (
+                    <ResultPanel
+                        header={apiError || 'State updated successfully'}
+                        primaryBtn={{
+                            text: 'Close',
+                            onClick: hideModal,
+                        }}
+                        iconClass={apiError ? 'fa-times-circle-o result-panel-icon--error' : ''}
                     />
```
avas27JTG, line 117:
```suggestion
                {showResultPanel && (
                    <ResultPanel
                        header={apiError || 'State updated successfully'}
                        primaryBtn={{
                            text: 'Close',
                            onClick: hideModal,
                        }}
                        iconClass={apiError ? 'fa-times-circle-o result-panel-icon--error' }
                    />
```

webapp/src/containers/updateState/index.tsx
manojmalik20: Adding a TODO comment.

webapp/src/containers/updateState/index.tsx
manojmalik20: What's the benefit of using `&&` instead of conditional operators `?:` here?

avas27JTG: it returns RHS if LHS holds true so in such cases where we just need to return only one thing and there is no else part, but `?:` is useful when there exists an else part 

here actually there is no else part we have just passed `""` because of `:`

manojmalik20: I think you're seeing the wrong thing. Please expand the code and then see. Here exists an `else` part.

webapp/src/containers/updateState/index.tsx
```diff
@@ -1,84 +1,141 @@
...
-                        required={true}
+                {showLoader && <CircularLoader/>}
+                {showResultPanel ? (
+                    <ResultPanel
+                        header={apiError || 'State updated successfully'}
+                        primaryBtn={{
+                            text: 'Close',
+                            onClick: hideModal,
+                        }}
+                        iconClass={apiError ? 'fa-times-circle-o result-panel-icon--error' : ''}
```
avas27JTG, line 117:
```suggestion
                        iconClass={apiError && 'fa-times-circle-o result-panel-icon--error'}
```

webapp/src/containers/updateState/index.tsx
avas27JTG: my bad, I meant this line https://github.com/Brightscout/mattermost-plugin-servicenow/pull/86/files#r1012845307

avas27JTG (approved):

PR #87, opened 2022-10-20
Bump plugin version to 1.2.0

ayusht2810:


manojmalik20 (approved):

webapp/src/containers/Rhs/rhs.scss
```diff
@@ -46,17 +46,16 @@
 }
 
 .share-record-btn {
-  align-items: center;
+  float: right;
```
avas27JTG, line 49: we should not use `float` it's deprecated now and not preferred

manojmalik20 (approved):

PR #90, opened 2022-10-21
[MI-2285] Add test cases for util.go and kv.go

ayusht2810:


server/plugin/kv_test.go
```diff
@@ -82,3 +85,104 @@ func TestStoreUser(t *testing.T) {
 		})
 	}
 }
+
+func TestDeleteUser(t *testing.T) {
+	defer monkey.UnpatchAll()
+	s := new(pluginStore)
```
Kshitij-Katiyar, line 91:
```suggestion
	ps := new(pluginStore)
```

server/plugin/kv_test.go
```diff
@@ -82,3 +85,104 @@ func TestStoreUser(t *testing.T) {
...
+	p := Plugin{}
+	s.userKV = kvstore.NewHashedKeyStore(kvstore.NewPluginStore(p.API), UserKeyPrefix)
+	for _, test := range []struct {
+		description   string
+		setupTest     func()
+		expectedError string
+	}{
+		{
+			description: "User is not loaded successfully from the KV store using mattermostID",
```
Kshitij-Katiyar, line 100: Is there any need for adding successfully in every test cases?
```suggestion
			description: "User is not loaded from the KV store using mattermostUserID",
```

server/plugin/kv_test.go
```diff
@@ -82,3 +85,104 @@ func TestStoreUser(t *testing.T) {
...
+			description: "User is not deleted successfully",
+			setupTest: func() {
+				monkey.PatchInstanceMethod(reflect.TypeOf(s), "LoadUser", func(*pluginStore, string) (*serializer.User, error) {
+					return testutils.GetSerializerUser(), nil
+				})
+				monkey.PatchInstanceMethod(reflect.TypeOf(s.userKV), "Delete", func(*kvstore.HashedKeyStore, string) error {
+					return errors.New("mockError")
+				})
+			},
+			expectedError: "mockError",
```
Kshitij-Katiyar, line 129: we can have this variable of type error and below we can check for equal values in this you don't have to convert the actual error to string

server/plugin/util_test.go
```diff
@@ -0,0 +1,202 @@
...
+		expectedResult string
+	}{
+		{
+			description: "ParseSubscriptionsToCommandResponse",
+			subscripitons: []*serializer.SubscriptionResponse{
+				{
+					SysID:              "mockID",
+					Type:               constants.SubscriptionTypeRecord,
+					Number:             "mockNumber",
+					ChannelID:          "mockChannelID",
```
Kshitij-Katiyar, line 31: create constants for variables which are same for all test cases

Kshitij-Katiyar (approved):

server/plugin/kv_test.go
```diff
@@ -74,11 +76,112 @@ func TestStoreUser(t *testing.T) {
...
+	for _, test := range []struct {
+		description   string
+		setupTest     func()
+		expectedError error
+	}{
+		{
+			description: "User is not loaded from the KV store using mattermostUserID",
+			setupTest: func() {
+				monkey.PatchInstanceMethod(reflect.TypeOf(ps), "LoadUser", func(*pluginStore, string) (*serializer.User, error) {
+					return nil, fmt.Errorf("mockError")
```
raghavaggarwal2308, line 102:
```suggestion
					return nil, fmt.Errorf("mockError")
```
Use something meaningful instead of "mockError". check everywhere

raghavaggarwal2308 (changes requested):

server/plugin/kv_test.go
ayusht2810: We have used this at many places and in other plugins as well. I think it's fine to use it

raghavaggarwal2308: It was removed from VA. If it doesn't take much time we can fix it here as well. Otherwise this also fine.

ayusht2810: should we create a ticket for this, and update this after all test cases are merged @manojmalik20 ?

manojmalik20: Yes @ayusht2810 You can do this. Please create a ticket.

server/plugin/util_test.go
```diff
@@ -0,0 +1,310 @@
...
+			assert.EqualValues(testCase.expectedResult, resp)
+		})
+	}
+}
+
+func TestIsAuthorizedSysAdmin(t *testing.T) {
+	defer monkey.UnpatchAll()
+	for _, testCase := range []struct {
+		description string
+		setupAPI    func(api *plugintest.API) *plugintest.API
```
manojmalik20, line 72:
```suggestion
		setupAPI    func(api *plugintest.API)
```

server/plugin/util_test.go
```diff
@@ -0,0 +1,310 @@
...
+		expectedErr bool
+		isAdmin     bool
+	}{
+		{
+			description: "IsAuthorizedSysAdmin: with admin",
+			setupAPI: func(api *plugintest.API) *plugintest.API {
+				api.On("GetUser", testutils.GetID()).Return(&model.User{
```
manojmalik20, line 79: You should sync this branch with `master`. There is a `GetUser` function in the testutils, you should use that here after modifying it.

server/plugin/util_test.go
```diff
@@ -0,0 +1,310 @@
...
+		})
+	}
+}
+
+func TestHandleClientError(t *testing.T) {
+	defer monkey.UnpatchAll()
+	requestURL := fmt.Sprintf("%s%s", constants.PathPrefix, constants.PathCreateSubscription)
+	for _, testCase := range []struct {
+		description        string
+		setupAPI           func(api *plugintest.API) *plugintest.API
```
manojmalik20, line 219:
```suggestion
		setupAPI           func(api *plugintest.API)
```

server/plugin/util_test.go
```diff
@@ -0,0 +1,310 @@
...
+			description: "handleClientError: with subscriptions not configured",
+			setupAPI: func(api *plugintest.API) *plugintest.API {
+				return api
+			},
+			setupPlugin:        func() {},
+			errorMessage:       errors.New(constants.APIErrorIDSubscriptionsNotConfigured),
+			expectedStatusCode: http.StatusBadRequest,
+		},
+		{
+			description: "handleClientError: with update set not uploaded",
```
manojmalik20, line 275:
```suggestion
			description: "handleClientError: with latest update set not uploaded",
```

server/plugin/util_test.go
```diff
@@ -0,0 +1,310 @@
...
+		},
+		{
+			description: "handleClientError: with subscriptions not authorized",
+			setupAPI: func(api *plugintest.API) *plugintest.API {
+				return api
+			},
+			setupPlugin:        func() {},
+			errorMessage:       errors.New(constants.APIErrorIDSubscriptionsNotAuthorized),
+			expectedStatusCode: http.StatusUnauthorized,
+		},
```
manojmalik20, line 292: There is one more condition now if you take pull from the latest `master`. Please write the test case for that also.

server/plugin/kv_test.go
ayusht2810: https://brightscout.atlassian.net/browse/MI-2351?atlOrigin=eyJpIjoiMDk1OTBmMDU0NDU3NDlkM2FjZTJkNTRkZjYzMGY2ZTEiLCJwIjoiaiJ9

created ticket for this

manojmalik20 (approved):

raghavaggarwal2308 (approved):

PR #92, opened 2022-10-27
[MI-2289] Added handling of refresh token expiry and other client errors

manojmalik20:
Created a util function handleClientError used by both APIs and slash commands
Changed the APIs and slash commands to use the new function
Created some new constants

### DIFF
--- a/server/plugin/command.go
+++ b/server/plugin/command.go
@@ -47,7 +47,6 @@ After that, this user will have the permission to add or manage subscriptions fr
 	helpCommandHeader                       = "#### Mattermost ServiceNow Plugin - Slash Command Help\n"
 	disconnectErrorMessage                  = "Something went wrong. Not able to disconnect user. Check server logs for errors."
 	disconnectSuccessMessage                = "Disconnected your ServiceNow account."
-	listSubscriptionsErrorMessage           = "Something went wrong. Not able to list subscriptions. Check server logs for errors."
 	listSubscriptionsWaitMessage            = "Your subscriptions will be listed soon. Please wait."
 	genericWaitMessage                      = "Your request is being processed. Please wait."
 	deleteSubscriptionErrorMessage          = "Something went wrong. Not able to delete subscription. Check server logs for errors."
--- a/server/plugin/command.go
+++ b/server/plugin/command.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"regexp"
-	"strings"
 	"sync"
 	"unicode"
 
@@ -56,6 +55,7 @@ After that, this user will have the permission to add or manage subscriptions fr
 	genericErrorMessage                     = "Something went wrong."
 	invalidSubscriptionIDMessage            = "Invalid subscription ID."
 	notConnectedMessage                     = "You are not connected to ServiceNow.\n[Click here to link your ServiceNow account.](%s%s)"
+	tokenExpiredReconnectMessage            = "Your connection with ServiceNow has expired, so you need to reconnect your account.\n[Click here to link your ServiceNow account.](%s%s)"
 	subscriptionsNotConfiguredError         = "It seems that subscriptions for ServiceNow have not been configured properly."
 	subscriptionsNotConfiguredErrorForUser  = subscriptionsNotConfiguredError + " Please contact your system administrator to configure the subscriptions by following the instructions given by the plugin."
 	subscriptionsNotConfiguredErrorForAdmin = subscriptionsNotConfiguredError + " To enable subscriptions, you have to download the update set provided by the plugin and upload that in ServiceNow. The update set is available in the plugin configuration settings. The instructions for uploading the update set are available in the plugin's documentation and also can be viewed by running the \"/servicenow help\" command."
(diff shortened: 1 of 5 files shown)


server/plugin/util.go
```diff
@@ -193,3 +194,75 @@ func filterSubscriptionsOnRecordData(subscripitons []*serializer.SubscriptionRes
 
 	return subscripitons[:n]
 }
+
+func (p *Plugin) handleClientError(err error, w http.ResponseWriter, r *http.Request, userID string, isSysAdmin bool, statusCode int, response string) string {
```
Kshitij-Katiyar, line 198: I think its better to put 2 string arguments together and put the error at end
```suggestion
func (p *Plugin) handleClientError(w http.ResponseWriter, r *http.Request, userID, response string, isSysAdmin bool, statusCode int, err error) string {  
```

server/plugin/util.go
```diff
@@ -193,3 +194,75 @@ func filterSubscriptionsOnRecordData(subscripitons []*serializer.SubscriptionRes
...
+			p.API.LogError(disconnectErr.Error())
+			if w != nil {
+				p.handleAPIError(w, &serializer.APIErrorResponse{StatusCode: http.StatusInternalServerError, Message: genericErrorMessage})
+			}
+			return genericErrorMessage
+		}
+
+		if w != nil {
+			p.handleAPIError(w, &serializer.APIErrorResponse{ID: constants.APIErrorIDRefreshTokenExpired, StatusCode: http.StatusBadRequest, Message: constants.APIErrorRefreshTokenExpired})
+			return message
```
Kshitij-Katiyar, line 215: in this case we are returning an empty string as message haven't been updated is this correct ?

server/plugin/util.go
```diff
@@ -193,3 +194,75 @@ func filterSubscriptionsOnRecordData(subscripitons []*serializer.SubscriptionRes
...
+			message = subscriptionsNotConfiguredErrorForAdmin
+		}
+
+		return message
+	}
+
+	if strings.EqualFold(err.Error(), constants.APIErrorIDSubscriptionsNotAuthorized) {
+		if w != nil {
+			p.handleAPIError(w, &serializer.APIErrorResponse{ID: constants.APIErrorIDSubscriptionsNotAuthorized, StatusCode: http.StatusUnauthorized, Message: constants.APIErrorSubscriptionsNotAuthorized})
+			return message
```
Kshitij-Katiyar, line 238: here as well

server/plugin/util.go
```diff
@@ -193,3 +194,75 @@ func filterSubscriptionsOnRecordData(subscripitons []*serializer.SubscriptionRes
...
+	if strings.EqualFold(err.Error(), constants.APIErrorIDLatestUpdateSetNotUploaded) {
+		if w != nil {
+			p.handleAPIError(w, &serializer.APIErrorResponse{ID: constants.APIErrorIDLatestUpdateSetNotUploaded, StatusCode: http.StatusBadRequest, Message: constants.APIErrorLatestUpdateSetNotUploaded})
+		}
+
+		return constants.APIErrorIDLatestUpdateSetNotUploaded
+	}
+
+	if w != nil {
+		if statusCode == 0 {
```
Kshitij-Katiyar, line 258: I think for some function calls we are not sending or either sending 0 as status code and for all those cases we are returning an internal server error then we can directly send 500  as status code from all those functions calls

server/plugin/command.go
```diff
@@ -48,14 +47,14 @@ After that, this user will have the permission to add or manage subscriptions fr
...
 	disconnectSuccessMessage                = "Disconnected your ServiceNow account."
-	listSubscriptionsErrorMessage           = "Something went wrong. Not able to list subscriptions. Check server logs for errors."
 	listSubscriptionsWaitMessage            = "Your subscriptions will be listed soon. Please wait."
 	genericWaitMessage                      = "Your request is being processed. Please wait."
 	deleteSubscriptionErrorMessage          = "Something went wrong. Not able to delete subscription. Check server logs for errors."
 	deleteSubscriptionSuccessMessage        = "Subscription successfully deleted."
 	genericErrorMessage                     = "Something went wrong."
 	invalidSubscriptionIDMessage            = "Invalid subscription ID."
 	notConnectedMessage                     = "You are not connected to ServiceNow.\n[Click here to link your ServiceNow account.](%s%s)"
+	tokenExpiredReconnectMessage            = "Your connection with ServiceNow has expired, so you need to reconnect your account.\n[Click here to link your ServiceNow account.](%s%s)"
```
ayusht2810, line 57: same

server/plugin/util.go
manojmalik20: Correct.

server/plugin/util.go
manojmalik20: Yes

server/plugin/util.go
manojmalik20: Yes, we can do that, but that will require more effort than this. This util is created for minimizing the caller's effort.

server/plugin/command.go
manojmalik20: Already updated in PR #99

server/plugin/util.go
Kshitij-Katiyar: ohh ok for me it looked weird equating the status code to 0 that's why

Kshitij-Katiyar (approved):

server/plugin/util.go
manojmalik20: That is the default value for integer type, that's why we are passing 0.

avas27JTG (approved):

PR #100, opened 2022-11-01
[MI-2313] Added unit tests for reducers

manojmalik20:
Changed extensions of files from tsx to ts


Kshitij-Katiyar (approved):

raghavaggarwal2308 (approved):

sooraj-shukla (approved):

webapp/src/reducers/updateStateModal/index.test.ts
```diff
@@ -0,0 +1,24 @@
+import {RecordType} from 'plugin_constants';
+
+import reducer, {showModal, hideModal, UpdateStateModalState} from 'reducers/updateStateModal';
+
+const previousState: UpdateStateModalState = {
```
avas27JTG, line 5: reading `UpdateStateModalState` is not sounding good, we should make it more descriptive mentioning what state it is updating like `UpdateDummyStateModalState`

manojmalik20: The name of the modal is "Update State", that's why this name was chosen. Is this better - `UpdateStateModalReduxState`?

avas27JTG (approved):

Issue #115, opened 2022-11-15
"Invalid number of params for this command" is too technical


DHaussermann:
Can we replace the text "Invalid number of params for this command" with something more clear and intuitive to all users?

![image](https://user-images.githubusercontent.com/35588783/202035423-55e13579-fdfd-43e8-bf25-232de2f2c777.png)

This could be replaced with something like "One or more fields is missing to run the command. Please run `/servicenow help` for more information"

ayusht2810:
This issue is fixed in the latest release v2.3.1. Please confirm and close this issue. @DHaussermann

DHaussermann:
Thanks @ayusht2810 confirmed fixed on 2.3.1.

Closing.

Issue #120, opened 2022-11-16
Several Modals seem to randomly be in the open state after the webapp has been left idle 


DHaussermann:
I'm seeing an odd behavior with Service Now. After I have left my local Mattermost instance idle for some time (maybe 2 hours or so), when I return to the browser tab, I see several modals open and their overlays showing for no apparent reason. 

https://user-images.githubusercontent.com/35588783/202306173-71dbc92f-12d7-4745-a749-7e3e8811af3c.mov

I'm working to see if there is any action that is causing this.     

![image](https://user-images.githubusercontent.com/35588783/202306335-d729e6f3-41cb-42fb-a388-8baa1970655e.png)
![image](https://user-images.githubusercontent.com/35588783/202306377-105abaab-32eb-4fac-ae04-33d52c0e58d4.png)
![image](https://user-images.githubusercontent.com/35588783/202306482-06a9b64b-bf4d-4b02-8e7c-95613496b3fa.png)
![image](https://user-images.githubusercontent.com/35588783/202306533-2b540360-5c36-40a0-93a0-89e272a60ba8.png)

In the image set above you can see that 4 modals where open. I have placed them chronologically

ayusht2810:
This issue is fixed in the latest release v2.3.1. Please confirm and close this issue. @DHaussermann

DHaussermann:
@ayusht2810 Are you certain this is resolved?
![image](https://user-images.githubusercontent.com/35588783/204658945-b72a59df-c6bb-4deb-845b-7628831b27e3.png)

This seems like something we should fix before release if possible. 

Note @mkdbns you can reach out to me if we need more concrete repro steps. This does not seem to happen with several modal as it did before but I can repro it for 1 modal in the other client.

hanzei:
Does that mean this bug only happens when a user is connected with more the one client?

ayusht2810:
@DHaussermann @hanzei should we proceed with this approach?

hanzei:
Sound good to me. But I would consider it an edge case. Other plugins suffer the same issue, e.g. GitHub.

@asatkinson What do you think?

asatkinson:
I think we decided that this was worth fixing, but not worth holding release up over.  For Phase 3 this will be fixed.

Issue #117, opened 2022-11-15
The Filters UX could be improved


DHaussermann:
Some suggestions for improving the UX on filters:
![image](https://user-images.githubusercontent.com/35588783/202038639-6c324678-0bd6-445d-b785-b88a3aed35bb.png)

1. If we can't save the sate of the filters per user - It may make more sense to have some type of `Show All` option visible. It would not have been obvious to me without reading the icon expose an option to see more data. 
2. Can the search option show at the top and push the top record down? It's odd that I can adjust the query while the top record remains hidden.
3. The use of the term `Hide` may be confusing in a UI that adjusts filters for data. Can we just switch it to `Close` or something else?

DHaussermann:
@manojmalik20 sorry to simplify point 1....
**I think**, the `Show all subscriptions` option is very valuable and it's not obvious at first look that it will be in there when I click the filter button. 0/5 maybe just curious if this can be made more discoverable?

DHaussermann:
Thanks @manojmalik20  confirmed fixed on 2.3.1.

3. has been resolved. We no no longer use the term `hide` to close a UI about filtering.

Point 1. is very subjective and 2. is minor. I'll follow up about maybe adding UX enhancements in future releases.

Closing.

PR #122, opened 2022-11-18
[MI-2372] Fixed issue #115, #117, #120 and add some enhancements

manojmalik20:
Issue #115 
Issue #117 (updated filter close button)
Issue #120 
Changed the structure of webapp by removing all the reducers for modals and creating one reducer for global modal Modified all the required containers and components to use the new global modal state Made some new types and refactored all the types to move to new files


Kshitij-Katiyar (approved):

webapp/src/websocket/index.ts
```diff
@@ -78,7 +70,7 @@ export function handleOpenCommentModal(store: Store<GlobalState, Action<Record<s
             recordType: data.record_type as RecordType,
             recordId: data.record_id,
         };
-        store.dispatch(showCommentModal(commentModalData) as Action);
+        store.dispatch(setGlobalModalState({modalId: 'addOrViewComments', data: commentModalData}) as Action);
```
avas27JTG, line 73: instead of hard-coded string values use constants for `modalId` as the same is used at least two places

ayusht2810: there is a TODO comment present for this
https://github.com/Brightscout/mattermost-plugin-servicenow/blob/4f01427b6e6fb4b1c33d1998440952aaeecf8155/webapp/src/types/common/index.d.ts#L5

avas27JTG (approved):

PR #125, opened 2022-11-24
[MI-2364] Create incident modal and add slash command to open incident modal

ayusht2810:
image
image
image


server/plugin/command.go
```diff
@@ -220,6 +220,22 @@ func (p *Plugin) handleSubscriptions(c *plugin.Context, args *model.CommandArgs,
...
+func (p *Plugin) handleCreate(c *plugin.Context, args *model.CommandArgs, parameters []string, client Client, isSysAdmin bool) string {
+	if len(parameters) == 0 {
+		return "Invalid create command. Available command is 'incident'."
+	}
+
+	command := parameters[0]
+	parameters = parameters[1:]
+
+	switch command {
+	case constants.SubCommandIncident:
```
Kshitij-Katiyar, line 232: do we need switch statement for 1 case only.

webapp/src/containers/createIncident/callerPanel.tsx
```diff
@@ -0,0 +1,111 @@
+import React, {forwardRef, useEffect, useState} from 'react';
+import {AutoSuggest} from '@brightscout/mattermost-ui-library';
+
+import Constants from 'src/plugin_constants';
+
+const incidentCallerOptions: CallerData[] = [
+    {
+        mattermostUserID: 'user id 1',
```
Kshitij-Katiyar, line 8: Ids are usually 1 words is there a reason here you have 3 words for it

webapp/src/containers/createIncident/callerPanel.tsx
```diff
@@ -0,0 +1,111 @@
...
+}
+
+const CallerPanel = forwardRef<HTMLDivElement, CallerPanelProps>(({
+    className,
+    actionBtnDisabled,
+    caller,
+    setCaller,
+    placeholder,
+}: CallerPanelProps): JSX.Element => {
+    const [callerOptions, setCallerlOptions] = useState<CallerData[]>(incidentCallerOptions);
```
Kshitij-Katiyar, line 61: I think this whole component is related to caller so there is no need to include caller in every state you are defining like callerAutoSuggestValue etc

webapp/src/containers/createIncident/styles.scss
```diff
@@ -0,0 +1,56 @@
+.incident-body {
```
Kshitij-Katiyar, line 1: get them reviewed by a FE developer or Abhishek

server/plugin/command.go
```diff
@@ -240,6 +256,16 @@ func (p *Plugin) handleSearchAndShare(_ *plugin.Context, args *model.CommandArgs
 	return ""
 }
 
+func (p *Plugin) handleCreateIncident(_ *plugin.Context, args *model.CommandArgs, params []string, client Client, _ bool) string {
```
raghavaggarwal2308, line 259: Just curious. If we don't need the first and last argument, why not remove them?

webapp/src/containers/createIncident/styles.scss
```diff
@@ -0,0 +1,56 @@
...
+                border-color: rgba(var(--center-channel-color-rgb), 0.2);
+            }
+        }
+
+    }
+
+    &__auto-suggest {
+        .auto-suggest {
+            width: 534px;
+            margin-left: -12px;
```
raghavaggarwal2308, line 53: Avoid negative values

raghavaggarwal2308 (approved):

server/plugin/command.go
ayusht2810: there will be one more command added later

webapp/src/containers/createIncident/callerPanel.tsx
ayusht2810: this is just mock data, it is removed after the integration

server/plugin/command.go
ayusht2810: I think we have done this to maintain consistency with other functions because in some they are being used

Kshitij-Katiyar (approved):

server/plugin/command.go
raghavaggarwal2308: Okay!...I don't think we should do that. 
Your opinion? @avas27JTG @manojmalik20

raghavaggarwal2308 (approved):

server/plugin/command.go
avas27JTG: I think we can omit the params that we don't need in a `func` as this will reduce code as well as passing extra arguments when calling it
If there is no other reason then I think we should maintain consistency only for design patterns and not for definition as that can vary as per need

webapp/src/containers/createIncident/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+    }, []);
+
+    const onShortDescriptionChangeHandle = (e: React.ChangeEvent<HTMLInputElement>) => {
+        setShortDescription(e.target.value);
+        setValidationError('');
+    };
+
+    const onDescriptionChangeHandle = (e: React.ChangeEvent<HTMLTextAreaElement>) => {
+        setDescription(e.target.value);
+    };
```
avas27JTG, line 77:
```suggestion
    const onDescriptionChangeHandle = (e: React.ChangeEvent<HTMLTextAreaElement>) => setDescription(e.target.value);
```

webapp/src/containers/createIncident/styles.scss
```diff
@@ -0,0 +1,56 @@
+.incident-body {
+    padding: 25px 32px 0px 32px;
```
avas27JTG, line 2:
```suggestion
    padding: 25px 32px 0 32px;
```
no need to put any unit for `0` in CSS

webapp/src/plugin_constants/index.ts
```diff
@@ -220,6 +220,21 @@ export const SubscriptionEventLabels: Record<SubscriptionEvents, string> = {
     [SubscriptionEvents.ASSIGNMENT_GROUP]: 'Assignment group changed',
 };
 
+export const IncidentImpactAndUrgencyOptions: DropdownOptionType[] = [
+    {
+        value: '1',
+        label: 'High',
```
avas27JTG, line 226: add `TODO` to remove all such mock data

webapp/src/selectors/index.ts
```diff
@@ -21,3 +21,7 @@ export const isCommentModalOpen = (state: PluginState): boolean => {
 export const isUpdateStateModalOpen = (state: PluginState): boolean => {
     return state.globalModalReducer.modalId === 'updateState';
 };
+
+export const isCreateIncidentModalOpen = (state: PluginState): boolean => {
+    return state.globalModalReducer.modalId === 'createIncident';
+};
```
avas27JTG, line 27:
```suggestion
export const isCreateIncidentModalOpen = (state: PluginState): boolean => state.globalModalReducer.modalId === 'createIncident';
```

webapp/src/plugin_constants/index.ts
ayusht2810: this is not mock data. Added `TODO` comments on the mock data

server/plugin/command.go
ayusht2810: added a TODO comment here, we can remove them after all commands are merged/added

ayusht2810:
closing this PR as it is already merged https://github.com/Brightscout/mattermost-plugin-servicenow/pull/1

PR #127, opened 2022-11-28
[MI-2394] Add feature to open and fill incident modal fields from pos…

ayusht2810:
…t menu options

image

### DIFF
--- a/webapp/src/containers/createIncident/createIncidentMenu.tsx
+++ b/webapp/src/containers/createIncident/createIncidentMenu.tsx
@@ -1,0 +1,70 @@
+import React, {MouseEvent, useCallback, useEffect} from 'react';
+import {useDispatch, useSelector} from 'react-redux';
+import {Action} from 'redux';
+
+import {getPost} from 'mattermost-redux/selectors/entities/posts';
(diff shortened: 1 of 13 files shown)


webapp/src/containers/createIncident/createIncidentMenu.tsx
```diff
@@ -0,0 +1,70 @@
...
+    useEffect(() => {
+        makeApiRequest(Constants.pluginApiServiceConfigs.getConnectedUser.apiServiceName);
+    }, []);
+
+    if (!show) {
+        return null;
+    }
+
+    return (
+        <>
```
Kshitij-Katiyar, line 49: there is only 1 <li> element so is this needed. please tell me if it is .

Kshitij-Katiyar (approved):

webapp/src/containers/createIncident/createIncidentMenu.tsx
ayusht2810: no, we don't need that (updated)

ayusht2810:
closing this PR as it is already merged https://github.com/Brightscout/mattermost-plugin-servicenow/pull/3

PR #132, opened 2022-11-30
[MI-2414] Show rhs header when subscriptions are not configured/autho…

ayusht2810:
…rized

image
image
image


Kshitij-Katiyar (approved):

webapp/src/containers/Rhs/rhs.scss
```diff
@@ -114,6 +114,11 @@
   width: 88px;
   height: 40px;
   border-radius: 4px;
+
+  &__without-filter {
+    position: absolute;
```
manojmalik20, line 119: Please get this reviewed by a FE as I don't think `position absolute` is the best way to do this.

manojmalik20 (approved):

raghavaggarwal2308 (approved):

webapp/src/containers/Rhs/rhs.scss
```diff
@@ -114,6 +114,11 @@
...
+    right: 0;
+  }
```
avas27JTG, line 121:
```suggestion
  &__without-filter {
    margin-left: auto;
  }
```

on line 42 we have `div className='d-flex align-item-center'` which means we have a flex container so to align any element inside a flex container to the rightmost you can use `margin-left: auto`

avas27JTG (approved):

webapp/src/containers/Rhs/rhs.scss
```diff
@@ -114,9 +114,25 @@
...
 
 .share-record-icon {
   margin-right: 3px;
   margin-top: 4px;
 }
+
+.configuration-not-enabled-err-state {
+  overflow: auto;
+
+  .flex-column {
```
avas27JTG, line 131: use other class of this div or add your own as `flex-column` is an util class and adding styles into it will not be good

ayusht2810: added a TODO comment

PR #137, opened 2022-12-02
Release to Marketplace

hanzei:
The sole purpose of this PR is to allow a review of the codebase of the plugin. It contains all changes that have been made since the repo was forked from https://github.com/mattermost/mattermost-plugin-starter-template.

The PR should not get merged. Once every party has approved the PR, https://github.com/mattermost/mattermost-plugin-servicenow/pull/138 will include the required changes for a release.

Security review ticket: https://mattermost.atlassian.net/browse/SEC-2365.


levb (approved):
Reviewed the golang part, and superficially - the webapp. Looks thorough. 👍

server/plugin/http.go
```diff
@@ -0,0 +1,112 @@
...
+		updatedError := strings.ReplaceAll(err.Error(), c.plugin.getConfiguration().ServiceNowBaseURL, "")
+		if strings.Contains(err.Error(), "dial tcp") {
+			errorData := strings.Split(c.plugin.getConfiguration().ServiceNowBaseURL, ":")
+			if len(errorData) == 3 {
+				updatedError = strings.ReplaceAll(updatedError, errorData[2], "")
+			}
+		}
+
+		return nil, http.StatusInternalServerError, errors.New(updatedError)
+	}
```
srkgupta, line 82: The current approach to fix the issue #148 is incorrect as it just looks for a specific pattern in the error message and replaces few strings. This might not suppress all the correct errors when an internal service is indeed running and that service throws a different error. The recommended way to fix this issue is to handle the http connection error and return a standard error message indicating "Unable to make connection to the specified ServiceNow instance". We should not return back any other error messages. For debugging purpose (if required), you can print the `err.Error()` in the server logs.

ayusht2810:
```
if err != nil {
		if strings.Contains(err.Error(), "invalid character '<'") {
			return nil, http.StatusInternalServerError, ErrorContentTypeNotJSON
		}

		if strings.Contains(err.Error(), "connect: connection refused") {
			return nil, http.StatusInternalServerError, ErrorConnectionRefused
		}

		return nil, http.StatusInternalServerError, err
	}

```

@srkgupta are you suggesting something like this? Here we are checking the connection error on the basis of whether it contains `connect: connection refused` words in it and returning a generic error to handle the http connection error, otherwise we return an error as it is.

srkgupta: Hi @ayusht2810. No, we should not check for any string patterns. We should just return a standard error message if `err != nil`

We can print the err.Error() as log statements. But we should not return the actual error message as server response.

srkgupta (approved):
Thanks all. All the security issues identified are fixed now. The changes LGTM from security perspective.

hanzei:
Closing as every party has approved the PR

Issue #135, opened 2022-12-02
Release v2.3.3


hanzei:
- [x] https://github.com/mattermost/mattermost-plugin-servicenow/pull/134
- [x] https://github.com/mattermost/mattermost-plugin-servicenow/pull/136
- [x] https://github.com/mattermost/mattermost-plugin-servicenow/pull/137

PR #138, opened 2022-12-02
Release v2.3.3

hanzei:
#### Summary
First release to the Marketplace

#### Ticket Link
Part of https://github.com/mattermost/mattermost-plugin-servicenow/issues/135


plugin.json
```diff
@@ -58,13 +63,15 @@
                 "display_name": "Encryption Secret:",
                 "type": "generated",
                 "help_text": "The secret key used to encrypt and decrpyt OAuth token.",
+                "placeholder": "",
```
manojmalik20, line 66: What's the benefit of adding this?

manojmalik20 (approved):
LGTM.

plugin.json
hanzei: Per https://github.com/mattermost/mattermost-server/blob/43e26ccda2db693b10aa98cd5cf5d93c480abf1c/model/manifest.go#L79 `placeholder` is not marked as `omitempty` and hence should be present even when empty.

hanzei: Yes, the changes have to function impact. They are just for consistency with the JSON schema.

Issue #140, opened 2022-12-08
`/servicenow subscriptions add` is not opening the modal as expected


DHaussermann:
There seems to be an issue on some servers where the modal for adding a subscription will not open. Nothing happens when you run the slash command. I see nothing relevant in browser dev logs or console logs.

The modal does open with the Add New subscription button. 

https://user-images.githubusercontent.com/35588783/206378678-d6d5598d-951f-4384-9945-306b0b20e1ea.mov

- This occurs on a cloud server I have running `5.37.3` I do not see the same issue occur locally.
- Both servers are running the `v2.3.3` build from the version bump PR in this repo. 
- This occurs in all browsers and the desktop client.

ayusht2810:
@DHaussermann this issue will be fixed with issue #120

DHaussermann:
@ayusht2810 are you able to provide some context on why my other server is not affected? 

Was it possible to open a modal with the command on earlier builds before `v2.3.3`?

DHaussermann:
@ayusht2810 after discussing this issue it could be lower in priority however on the `5.37` server in the video - There is no button to add additional subscriptions beyond the 1st once 1 is created. 
![image](https://user-images.githubusercontent.com/35588783/206550695-d531fa08-7129-408c-a549-ffa0888505aa.png)

This also does not repro on my local. Ping me if it helps to send an invite to the affected server. Also happy to open a separate issue here if neded.

DHaussermann:
@ayusht2810 I also see this issue on the `Add` and `Update`  buttons can you see if this is the same probem 

![image](https://user-images.githubusercontent.com/35588783/207070619-321385b5-3235-4b90-9de7-23501e4a2680.png)

ayusht2810:
@DHaussermann this issue is fixed in the PR https://github.com/mattermost/mattermost-plugin-servicenow/pull/141

ayusht2810:
@DHaussermann These are fixed so please verify and close

ayusht2810:
@DHaussermann reminder to verify the issue.

raghavaggarwal2308:
@ayusht2810 Verified, the above issue are fixed now.

Tested on MM cloud server version v5.37.3 with master branch

PR #141, opened 2022-12-12
Fix issue 140

ayusht2810:
#### Summary

1. Fixed the issue of modals not opening on using the slash command.
2. Fixed the `Add Subscription` button visibility after adding a new subscription in RHS.
3. Fixed the `Created By` filter not working.

#### Ticket Link
[Issue #140](https://github.com/mattermost/mattermost-plugin-servicenow/issues/140)

### DIFF
--- a/webapp/src/containers/Rhs/rhs.scss
+++ b/webapp/src/containers/Rhs/rhs.scss
@@ -48,8 +48,11 @@
 }
 
 .rhs-btn-container {
-  @include positioning(fixed, "", 0, 0, 1px, 1);
-  padding-left: 11px;
+  position: absolute;
+  right: 0;
+  bottom: 0;
+  left: 1px;
+  z-index: 1;
   box-shadow: 9px 2px 5px 5px var(--center-channel-bg);
 }
 
--- a/webapp/src/containers/Rhs/index.tsx
+++ b/webapp/src/containers/Rhs/index.tsx
@@ -216,7 +216,7 @@ const Rhs = (): JSX.Element => {
                     showAllSubscriptions={showAllSubscriptions}
                     setShowAllSubscriptions={setShowAllSubscriptions}
                     filter={filter}
-                    setFilter={setFilter}
+                    setFilter={handleSetFilter}
                     setResetFilter={setResetFilter}
                 />
             )}
--- a/webapp/src/index.tsx
+++ b/webapp/src/index.tsx
@@ -43,7 +43,10 @@ export default class Plugin {
         registry.registerChannelHeaderButtonAction(<ServiceNowIcon className='servicenow-icon'/>, () => store.dispatch(toggleRHSPlugin), null, Constants.ChannelHeaderTooltipText);
         registry.registerAdminConsoleCustomSetting('ServiceNowUpdateSetDownload', DownloadButton);
         const iconUrl = `${Utils.getBaseUrls().publicFilesUrl}${Constants.SERVICENOW_ICON_URL}`;
-        registry.registerAppBarComponent(iconUrl, () => store.dispatch(toggleRHSPlugin), Constants.ChannelHeaderTooltipText);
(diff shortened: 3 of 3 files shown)


manojmalik20 (approved):

DHaussermann:
Thanks @ayusht2810  
I have taken a look and deployed this branch to an HA server. 
- All Modals mentioned in the issue are opening 
- Tested in browser and desktop app
- No issues found.

I'll watch fore this fix to be included in a release.

manojmalik20:
@DHaussermann Should we merge this PR?

DHaussermann:
@manojmalik20 based on the solution you've provided, do you no of any other risk that should be tested before hand? 
Let me know.

If not this would be ready to merge but, I'm not sure if you normally have 2 dev reviews or not.

webapp/src/containers/Rhs/rhs.scss
```diff
@@ -48,8 +48,11 @@
 }
 
 .rhs-btn-container {
-  @include positioning(fixed, "", 0, 0, 1px, 1);
-  padding-left: 11px;
+  position: absolute;
```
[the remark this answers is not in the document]
mickmister: Are we able to remove the `@brightscout/mattermost-ui-library/build/styles/abstract/_mixins` import since we've removed the one line referencing the mixins?

mickmister (changes requested):

webapp/src/containers/addOrEditSubscriptions/subComponents/index.tsx
```diff
@@ -213,7 +213,7 @@ const AddOrEditSubscription = ({open, close, subscriptionData}: AddOrEditSubscri
         const setHeight = (modalContent: Element) => modalContent.setAttribute('style', `height:${bodyHeight + PanelDefaultHeights.panelHeader}px`);
 
         // Select all the modal-content elements and set the height
-        document.querySelectorAll('.rhs-modal.add-edit-subscription-modal .modal-content').forEach((modalContent) => setHeight(modalContent));
+        document.querySelectorAll('.servicenow-rhs-modal.add-edit-subscription-modal .modal-content').forEach((modalContent) => setHeight(modalContent));
```
mickmister, line 216: Can this be done in a more React-friendly way? Accessing `document` directly usually isn't the best idea

webapp/src/containers/updateState/index.tsx
```diff
@@ -109,7 +109,7 @@ const UpdateState = () => {
         <Modal
             show={isUpdateStateModalOpen(pluginState)}
             onHide={hideModal}
-            className='rhs-modal'
+            className='servicenow-rhs-modal'
```
mickmister, line 112: What is an RHS modal? I would think the RHS and a modal are two different concepts

mickmister (approved):
Thanks for addressing the issues @ayusht2810 :+1:

mickmister:
Note that my comments above are not blocking, and this is good to merge from dev perspective cc @DHaussermann

DHaussermann (approved):
Tested and passed as per comment above 

LGTM!

Is there anything else we're waiting on her if not, Please merge so it can be included in the next build @ayusht2810.

ayusht2810:
We will fix the above review comments in a new PR. Merging this PR for now.

webapp/src/containers/addOrEditSubscriptions/subComponents/index.tsx
manojmalik20: @mickmister The React-friendly way to do this would be to use `ref` but that is not possible here as we are trying to access an element inside the React-Bootstrap modal. I tried to do this using ref but failed. Also, in this case, we are using `document` directly only to apply styling to the element and we are not modifying the DOM structure, so I think using `document` here is not a major issue. What do you think?

PR #171, opened 2023-02-17
Added a logic to disconnect users from ServiceNow on change of the encryption secret.

Nityanand13:
#### Summary

When an admin changes the encryption secret then all the users will be disconnected from ServiceNow and they have to reconnect their accounts to ServiceNow.


ayusht2810:
@hanzei the below PR is ready for review. Can please assign the designated reviewer to it.

server/plugin/configuration.go
```diff
@@ -141,5 +142,10 @@ func (p *Plugin) OnConfigurationChange() error {
 	configuration.PluginID = manifest.ID
 
 	p.setConfiguration(configuration)
+
+	if oldEncryptionSecret != "" && oldEncryptionSecret != p.getConfiguration().EncryptionSecret {
+		go p.store.DeleteUserTokenOnEncryptionSecretChange()
+	}
```
mickmister, line 148: Shouldn't we delete the users even if the `oldEncryptionSecret` is blank? As long as the secret changed

hanzei (changes requested):
Thanks for the PR :+1:

I'm wondering how the code behaves in a HA environment. Are there issues that can arise when multiple servers race to delete the user list?

AayushChaudhary0001 (approved):
The PR above has been tested for the following scenario:-

- On changing the encryption secret in MM, the connected users are disconnected from Servicenow.

The PR was working fine for the above condition, LGTM. Approved.

Kshitij-Katiyar:
> Thanks for the PR 👍
> 
> I'm wondering how the code behaves in a HA environment. Are there issues that can arise when multiple servers race to delete the user list?

@hanzei https://github.com/mattermost/mattermost-plugin-servicenow/pull/171#pullrequestreview-1945840706 QA tested this PR for the HA environment and its working perfectly fine.

hanzei:
I'm leaving it to @mickmister to suggest next steps on this PR

server/plugin/kv.go
```diff
@@ -80,34 +82,48 @@ func (s *pluginStore) DeleteUser(mattermostUserID string) error {
 func (s *pluginStore) GetAllUsers() ([]*serializer.IncidentCaller, error) {
 	page := 0
 	users := []*serializer.IncidentCaller{}
+	wg := new(sync.WaitGroup)
+	mu := new(sync.Mutex)
```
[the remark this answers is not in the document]
hanzei: > We definitely want to coordinate this so it only runs on one node. We can use [cluster.Mutex](https://github.com/mattermost/mattermost/blob/7e317c7568699ddf31a43b3177ccb04168e28b8f/server/public/pluginapi/cluster/mutex.go#L41) to coordinate this. We did something similar (a "migration" of sorts) using this here https://github.com/mattermost/mattermost-plugin-gitlab/pull/361.

Strong :+1:

server/plugin/kv.go
```diff
@@ -80,34 +82,48 @@ func (s *pluginStore) DeleteUser(mattermostUserID string) error {
...
 		for _, key := range kvList {
-			if userID, isValidUserKey := IsValidUserKey(key); isValidUserKey {
-				decodedKey, decodeErr := decodeKey(userID)
-				if decodeErr != nil {
-					s.plugin.API.LogError("Unable to decode key", "UserID", userID, "Error", decodeErr.Error())
-					continue
+			wg.Add(1)
+
+			go func(key string) {
+				defer wg.Done()
```
hanzei, line 97: Do we need to fan out the keys to different go routines as part of this PR?

server/plugin/kv.go
```diff
@@ -118,6 +134,21 @@ func (s *pluginStore) GetAllUsers() ([]*serializer.IncidentCaller, error) {
...
+func (s *pluginStore) DeleteUserTokenOnEncryptionSecretChange() {
+	users, err := s.GetAllUsers()
+	if err != nil {
+		s.plugin.API.LogError(constants.ErrorGetUsers, "Error", err.Error())
+		return
+	}
+
+	for _, user := range users {
+		if err := s.DeleteUser(user.MattermostUserID); err != nil {
+			s.plugin.API.LogError("Unable to delete a user", "UserID", user.MattermostUserID, "Error", err.Error())
```
hanzei, line 146:
```suggestion
			s.plugin.API.LogWarn("Unable to delete a user", "UserID", user.MattermostUserID, "Error", err.Error())
```

hanzei: The log message should make it clear why we are deleting the user.

hanzei (changes requested):

server/plugin/kv.go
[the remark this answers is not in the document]
hanzei: I wonder why we are fanning out the requests at all.

ayusht2810:
@hanzei @mickmister fixed the review comments. Please re-review.

ayusht2810:
@mickmister @hanzei Gentle reminder to re-review the PR

mickmister:
There's no customer reported ticket associated with this, so I'm not sure if this is a priority that should require this complexity

Kshitij-Katiyar:
> There's no customer reported ticket associated with this, so I'm not sure if this is a priority that should require this complexity

This was a problem we faced while developing, that's why we have added this feature as an enhancement. 
@mickmister

server/plugin/kv.go
```diff
@@ -115,9 +128,61 @@ func (s *pluginStore) GetAllUsers() ([]*serializer.IncidentCaller, error) {
...
+
+	deleteAllUsersMutex.Lock()
+	if deleteAllUsersState := s.DeleteAllUsersState(); !deleteAllUsersState {
+		deleteAllUsersMutex.Unlock()
+		return
+	}
+
+	defer func() {
+		if err = s.basicKV.Delete(constants.DeleteAllUsersKey); err != nil {
+			s.plugin.API.LogError("Unable to delete all users state from the store", "Error", err.Error())
```
mickmister, line 149: I'm thinking "delete all users state" may be a bit ambiguous. Maybe we can word this as "disconnect users job running flag". Same for other similar log statements

server/plugin/kv.go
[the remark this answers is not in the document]
mickmister: Yeah I'm thinking the same now. I think we should do this synchronously here with no wait group or goroutines. @Kshitij-Katiyar What do you think?

server/plugin/kv.go
```diff
@@ -80,33 +84,42 @@ func (s *pluginStore) DeleteUser(mattermostUserID string) error {
 func (s *pluginStore) GetAllUsers() ([]*serializer.IncidentCaller, error) {
 	page := 0
 	users := []*serializer.IncidentCaller{}
+	wg := new(sync.WaitGroup)
+	mu := new(sync.Mutex)
 	for {
+		wg.Add(1)
```
mickmister, line 90: The wait group is actually one per page here, so I misspoke above. I'm still thinking our own database should not be used like this in this context

mickmister (commented):
The PR mostly LGTM. I think the goroutines used to fetch all users may be unnecessary. Otherwise looks good :+1:

server/plugin/kv.go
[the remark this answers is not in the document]
Kshitij-Katiyar: yeah, I think we should do this synchronously as this is a very minor edge case as well. @mickmister

ayusht2810:
@mickmister fixed the review comments. Please re-review

server/plugin/kv.go
```diff
@@ -90,21 +95,24 @@ func (s *pluginStore) GetAllUsers() ([]*serializer.IncidentCaller, error) {
 			if userID, isValidUserKey := IsValidUserKey(key); isValidUserKey {
 				decodedKey, decodeErr := decodeKey(userID)
 				if decodeErr != nil {
-					s.plugin.API.LogError("Unable to decode key", "UserID", userID, "Error", decodeErr.Error())
-					continue
+					s.plugin.API.LogError("Unable to decode key", "Key", key, "Error", decodeErr.Error())
+					break
```
mickmister, line 99: If we reach an error, we should either `continue` or `return` here I think. The original strategy of `continue` seems fine to me

server/plugin/kv.go
```diff
@@ -90,21 +95,24 @@ func (s *pluginStore) GetAllUsers() ([]*serializer.IncidentCaller, error) {
...
 				}
 
+				// Append the loaded user to the users slice under a lock.
+				mu.Lock()
 				users = append(users, &serializer.IncidentCaller{
 					MattermostUserID: user.MattermostUserID,
 					Username:         user.Username,
 					ServiceNowUser:   user.ServiceNowUser,
 				})
+				mu.Unlock()
```
mickmister, line 115: We shouldn't need this lock anymore

mickmister (commented):
LGTM, just one suggestion to remove the lock since we're doing this synchronously now

ayusht2810:
@mickmister fixed the review comments. Please re-review

mickmister (approved):

server/plugin/kv.go
```diff
@@ -118,6 +121,58 @@ func (s *pluginStore) GetAllUsers() ([]*serializer.IncidentCaller, error) {
 	return users, nil
 }
 
+func (s *pluginStore) DeleteUserTokenOnEncryptionSecretChange() {
+	deleteAllUsersMutex, err := cluster.NewMutex(s.plugin.API, constants.DeleteAllUsersMutexKey)
+	if err != nil {
+		s.plugin.API.LogError("Failed to create mutex for deleting users", "Error", err.Error())
+	}
+
+	deleteAllUsersMutex.Lock()
```
fmartingr, line 130: Can't we move the mutex inside the `DeleteAllUsersState` since that's the method blocked on it? Is there a reason for having it here?

server/plugin/kv.go
```diff
@@ -118,6 +121,58 @@ func (s *pluginStore) GetAllUsers() ([]*serializer.IncidentCaller, error) {
...
+
+	for _, user := range users {
+		if err := s.DeleteUser(user.MattermostUserID); err != nil {
+			s.plugin.API.LogWarn("Unable to delete a user on encryption secret change", "UserID", user.MattermostUserID, "Error", err.Error())
+			continue
+		}
+	}
+}
+
+func (s *pluginStore) DeleteAllUsersState() bool {
```
fmartingr, line 157: I didn't understood what this did at first. I guess this is checking if the delete all users process is running, maybe a more descriptive name or comment is in order?

fmartingr (approved):
Good work. I needed a few reads to understood what was going on, left my thought in comments but not blocking.

This would be fairly useful for other plugins as well @mickmister (https://github.com/mattermost/mattermost-plugin-google-calendar/issues/61)

mickmister (approved):

arush-vashishtha (approved):
This PR has been tested for updating the "Encryption Secret", The connected users are getting disconnected. LGTM

PR #178, opened 2023-04-11
[MM-51955] Use mattermost-server/v6

hanzei:
#### Summary
Migrate codebase to use `mattermost-server/v6` and update min server version accordingly.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-51955


manojmalik20 (approved):

plugin.json
```diff
@@ -7,7 +7,7 @@
     "release_notes_url": "https://github.com/mattermost/mattermost-plugin-servicenow/releases/tag/v2.3.4",
     "icon_path": "assets/icon.svg",
     "version": "2.3.4",
-    "min_server_version": "5.37.0",
+    "min_server_version": "6.3.0",
```
trilopin, line 10: interesting, several questions here:

1. is this change required by the v6 migration? should I do it in the mscalendar v6 migration?
2. is there any standard procedure to communicate this? do we need approval for that?

trilopin (commented):
All code seems fine to me 🎉 

Just have questions for min_server clause

plugin.json
hanzei: 1. In general, it's discouraged to run a plugin which has been complied with a new code version against an older server version. Hence, I'm updating the `min_server_version` here.
2. I don't think so. The version increment does not affect existing plugin installations.

Issue #167, opened 2023-01-25
ServiceNow Slash Commands which launch dialogs do not run inside Playbooks


asatkinson:
If I run the slash command /servicenow create incident in the channel the slash command works fine and shows the dialog.

However if I run the same slash command inside a playbook the dialog does not open.

Dialogs used to open, but doesn't anymore.  

The playbooks team says the slash commands with dialogs they have tried work correctly.

I'm wondering if the change in how dialogs get launched might be causing an issue.  

I am running ServiceNow plugin 3.2.0, Mattermost 7.7.1
Here is a screenshot of two commands embedded in a playbook.  The first one should launch a dialog - but does not
The second one only sends content to the channel and works fine.

![image](https://user-images.githubusercontent.com/45946404/214611858-85c49f5b-75f3-4e5e-b45a-712234c3bdf8.png)

mickmister:
@asatkinson This is because the first command is implemented as webapp plugin. The playbooks plugin only runs server-side commands, so this ends up being a no-op for webapp plugin commands.

The alternative would be to use interactive dialogs. App framework commands are currently unsupported with Playbooks, but would be supported if webapp plugin ones were also supported. The App framework command parsing happens client-side atm, which is essentially the same issue with the webapp plugins.

mkdbns:
That is a good summary @mickmister. The team has been looking at some options. This is one we are investigating:

Option A:
* We can call an API when a slash command is posted using slashCommandWillBePostedHook having a random id and slash command in the request body. (need to explore how the Mattermost web app does this)
* We will store the random id generated above in the redux state.
* The API call on the server side will call a function to execute the slash command which will dispatch the necessary information and the random id to the client using a WebSocket event.
* The id received from the WebSocket event will be matched with the id present in the redux state to perform the required actions in a single client.

Here are others but require changes outside our plugin so not ideal:

Option B. Code changes in Mattermost server - Store the trigger id received in the command arguments without encrypting it so we can match that id at the client side by dispatching a WebSocket event.

Option C. Code changes in the playbook plugin - Update the code of the playbook plugin such that it somehow calls the slashCommandWillBePostedHook before the call of the API.

asatkinson:
@mickmister there is already a discussion in the CE Brightscout channel on this with the playbooks team.
@mkdbns is moving back to the approach used to launch the modal in the 2.X stream an option - at least temporarily while this is worked out?

ayusht2810:
[...]

PR #182, opened 2023-04-21
Fixed issue #172

### LINKED ISSUE #172: Modify the Create Incident Message Action Dialog and back-port it to the 2.X stream
**Background**
As part of the 3.X stream of work, functionality to create incidents and to subscribe to them was added.
Customer interviews have identified a desire to slightly change this behaviour.  We also would like to make this functionality available in the 2.X stream so that it can be quickly released without an extensive security review.

Currently, the "Message Action" for a Mattermost message looks like this.
image

We would like to make three small changes as described in the objectives so that this can be quickly released to the customer for evaluation.

**Objectives**

For the message actions dialog, 
1. Move the Mattermost message to the "short description" field.  If the message is too long, add a ... to this field.  Try several different cutoff lengths and choose which one works best in both Mattermost and in ServiceNow.
2. If the short description field extends past the "cutoff" length, put the full message in the description field (as today)
3. Identify the SENDER of the Mattermost Message and if they have a ServiceNow linked account, use this account to populate the "Caller" Field.  
4. Remove the Impact and urgency fields from the form

For the create new incident dialog
Leave this dialog as-is for now if possible.

**Key Results**
Demonstrate that the message actions dialog for ServiceNow properly populates the short description, description and caller fields and that the impact and urgency fields are removed.
Demonstrate that this runs correctly in both the 3.X and 2.X streams.

ayusht2810:
#### Summary

1. Added a feature to open the incident modal using the slash command and to create incidents using the slash command.
2. Added a feature in the incident modal to subscribe to the newly created incident.
3. Added a feature to open the incident modal from post menu actions and auto-fill the "Short description" and "Description" fields with the post data.

#### Screenshots
image
image

#### Ticket Link

Fixes #172


trilopin (changes requested):
Thanks for your work @ayusht2810 

here I write some minor things I noticed from a first review:


1. distance between buttons/labels in modal seems too small



Screenshot 2023-05-09 at 12 33 49
Screenshot 2023-05-09 at 12 38 57
Screenshot 2023-05-09 at 12 39 20
Screenshot 2023-05-09 at 12 41 07
Screenshot 2023-05-09 at 12 41 10



2. add comments modal has cta button enabled before adding context inside the textarea (update state is correct)

3. Update modal raises in multiple devices at the same time, not only on the window that was triggered (it doesn't happen with the create incident modal)

trilopin:
Additionally I get `It seems that you are not authorized to manage subscriptions in ServiceNow. Please follow the instructions for setting up user permissions available in the plugin's documentation. The instructions can also be viewed by running the "/servicenow help" command.` message when I try to raise the modal with `/servicenow create incident ` (it works fine for post dialog)

I could understand that I lack permissions for subscriptions, but should they apply to create-incident? if so, should it apply for create incident from post dialog too?

ayusht2810:
@trilopin regarding point
`3. Update-State modal raises in multiple devices at the same time, not only on the window that was triggered (it doesn't happen with the create incident modal)`. We fixed this issue previously in a PR but this caused the issue of `modals` not opening through the `Playbooks plugin`. You can refer to these issues for more detailed discussion.
https://github.com/mattermost/mattermost-plugin-servicenow/issues/120
https://github.com/mattermost/mattermost-plugin-servicenow/issues/167

trilopin:
@hanzei @mickmister

ayusht2810:
@trilopin fixed the review fixes

server/plugin/api.go
```diff
@@ -752,7 +751,18 @@ func (p *Plugin) createIncident(w http.ResponseWriter, r *http.Request) {
...
+	channelID := incident.ChannelID
+	if channelType == model.ChannelTypeDirect || channelType == model.ChannelTypeGroup {
+		channel, err := p.API.GetDirectChannel(userID, p.botID)
+		if err != nil {
+			p.API.LogError("Couldn't get bot's DM channel", "UserID", userID, "Error", err.Error())
+			return
+		}
+
+		channelID = channel.Id
+	}
```
lieut-data, line 763: Apologies that I'm not familiar with the overall feature here, but I'm wondering why we redirect the sharing post into the DM between the user and the bot in this case instead of just posting as the bot into the DM/GM?

It's true that the bot won't be a member of these channels, but fundamentally I don't think that prevents the bot from posting a message there. Not a blocker, but just sharing to make sure I understand the intent.

server/plugin/util.go
```diff
@@ -308,23 +308,23 @@ func decodeKey(key string) (string, error) {
 	return string(decodedKey), nil
 }
 
-func (p *Plugin) HasChannelPermissions(userID, channelID string) (int, error) {
+func (p *Plugin) HasChannelPermissions(userID, channelID string, checkType bool) (int, model.ChannelType, error) {
```
lieut-data, line 311: I'm guessing some of this is upstream, so I won't recommend a change here, but it would be ideal to avoid boolean parameters like this in favour of explicit methods instead, e.g. `HasChannelPermissions` and `HasPublicOrPrivateChannelPermissions`. It's much easier to review the calling code instead of an obscure `, true` or `, false` addition.

server/plugin/client.go
```diff
@@ -270,9 +270,13 @@ func (c *client) GetMe(userEmail string) (*serializer.ServiceNowUser, int, error
 }
 
 func (c *client) CreateIncident(incident *serializer.IncidentPayload) (*serializer.IncidentResponse, int, error) {
+	queryParams := url.Values{
+		constants.SysQueryParamDisplayValue: {"true"},
```
lieut-data, line 274: Can you clarify what this does?

lieut-data (approved):
Just a handful of comments, but nothing blocking.

webapp/src/containers/createIncident/createIncidentMenu.tsx
```diff
@@ -0,0 +1,65 @@
...
+                className='MenuItem'
+                role='menuitem'
+            >
+                <button
+                    className='incident-menu'
+                    role='presentation'
+                    onClick={handleClick}
+                >
+                    <img
+                        src={`${Utils.getBaseUrls().publicFilesUrl}${Constants.SERVICENOW_ICON_URL}`}
```
mickmister, line 54: We'll want to make sure we take into account the server's Site URL when contacting the plugin via HTTP. At the moment we are just parsing the browser's URL:

https://github.com/mattermost/mattermost-plugin-servicenow/blob/fefb220d1256c4e56d130beb54c30819196da598/webapp/src/utils/index.tsx#L18-L23

server/plugin/command.go
```diff
@@ -132,7 +132,7 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
 		}
 
 		var client Client
-		if action != constants.CommandDisconnect && action != constants.CommandSearchAndShare {
+		if action != constants.CommandDisconnect && action != constants.CommandSearchAndShare && action != constants.CommandIncident {
 			if client = p.GetClientFromUser(args, user); client == nil {
 				return &model.CommandResponse{}, nil
 			}
```
mickmister, line 138: Can we instead have this if statement be a "positive" one? Meaning, instead of adding exception cases, we check for the specific case(s) where we should run the if block. This way we don't need to update this condition when we add more features.

mickmister (approved):
Just a few non-blocking comments

DHaussermann:
HI @ayusht2810 
I have done a round of testing on this and it seems to be working. I still have some edge cases to explore and such. 

I do have a coupe questions with the business logic on the "caller" field. From what I can tell...
- When creating an incident not based on a post - the creator may assign any Service Now user as the caller. 
- When creating an incident based on a post of a user **not** connected to Service Now -  the creator may assign any Service Now user as the caller.  
- When creating an incident based on a post of a user who **is** connected to Service Now - The author of the post must be assigned as the caller
Is this ☝️ correct?

If so... 

1. Are we certain we want to enforce that the author of the post must be the caller if they are connected to Service Now? As it is, the short description can be edited from the post content and it may be the case that you want to create an incident based on another users post
2. When using slash command to create (or creating based on a post from a user not connected to SN) any user can be the caller. The problem is that both the Mattermost username and the ServiceNow username are shown. 
image
In this image - both users are the same person and called will be `abel.tutor` in SN regardless of which option I select. Can we add some kind of logic to hide Mattermost username if Service Now username is already included? I am worried this could cause alot of "duplication" and confusion.

3. When caller is locked to the author of the post, can the field be read-only? Even though there is not affect on the SN side, it could be confusing that they can attempt to change this value
image

DHaussermann:
Hi @ayusht2810 I have only a couple things to add from my test pass. 

4. Short description field should be capped at 160 (or 161) characters in the modal text field. On Service Now side the field only holds 161 characters. The remaining characters are silently discarded.
5. 0/5 minor UX point - The channel and user pickers are a bit odd as the styling is different from other plugins. But also, it's different because you must clear out the current selection to pick something else rather then seeing a drop down to make a change.

server/plugin/client.go
ayusht2810: this query param helps in retrieving the display values instead of their IDs for the fields

ayusht2810:
> HI @ayusht2810 I have done a round of testing on this and it seems to be working. I still have some edge cases to explore and such.
> 
> I do have a coupe questions with the business logic on the "caller" field. From what I can tell...
> 
> * When creating an incident not based on a post - the creator may assign any Service Now user as the caller.
> * When creating an incident based on a post of a user **not** connected to Service Now -  the creator may assign any Service Now user as the caller.
> * When creating an incident based on a post of a user who **is** connected to Service Now - The author of the post must be assigned as the caller
>   Is this point_up correct?

Yes, the above behavior is correct. But in the 3rd point, we are not restricting the caller. The caller field can be configured. Also, the reason for having this field configured is if the system admin wants to make an incident on behalf of some other user, he can change the caller value for that particular incident.

> If so...
> 
> 1. Are we certain we want to enforce that the author of the post must be the caller if they are connected to Service Now? As it is, the short description can be edited from the post content and it may be the case that you want to create an incident based on another users post
> 2. When using slash command to create (or creating based on a post from a user not connected to SN) any user can be the caller. The problem is that both the Mattermost username and the ServiceNow username are shown.
>    image
>    In this image - both users are the same person and called will be `abel.tutor` in SN regardless of which option I select. Can we add some kind of logic to hide Mattermost username if Service Now username is already included? I am worried this could cause alot of "duplication" and confusion.

We are not able to reproduce this issue. Can you please re-verify whether both the users are the same user or different one?

> 3. When caller is locked to the author of the post, can the field be read-only? Even though there is not affect on the SN side, it could be confusing that they can attempt to change this value
>    image

ayusht2810:
> 5. 0/5 minor UX point - The channel and user pickers are a bit odd as the styling is different from other plugins. But also, it's different because you must clear out the current selection to pick something else rather then seeing a drop down to make a change.

@DHaussermann we are using an `Auto Suggest` component here instead of a drop-down. That's why the behavior is like this.

raghavaggarwal2308:
@mickmister Fixed the review comments mentioned by you

DHaussermann:
On 1 . 
>"But in the 3rd point, we are not restricting the caller. The caller field can be configured."

Maybe I'm misunderstanding... But, when the author of the post is mapped to Service Now, I see no obvious way to change the value..

Mapped User
image


Unmapped User
image


On 2. I can no longer see the duplication but, only the Mattermost usernames are shown not the Service Now user names please confirm this is correct.

On 3. whats happening here is this control - It's actually unclear what is selected  and what text is just the search term. So, if I search for something that does not match anything, I can then submit with my search term and it's not obvious what is selected. This is what caused my confusion. Can you force me to select something?
image

DHaussermann:
@raghavaggarwal2308 I'm sorry the comment above ☝️  did not save yesterday but, I meant to @ mention you.

raghavaggarwal2308:
@DHaussermann 
1. I think there is a misunderstanding. Since this is a search component, not a dropdown, we must remove the typed value to see all available options. Currently, it's showing all the options matching the typed value.

2. Yes, it's correct, we are showing the mattermost usernames.

3. If the user does not select any value from the list and types any random value and submits then the default value will be empty(We are taking an assumption that it's obvious for the user to select a value and not just type it). Since this field is not mandatory for sending in the ServiceNow API, we are not forcing the user to select a value.

DHaussermann:
@raghavaggarwal2308 

1. Yes I was confused. Without the `x` I saw no obvious way to change the value (without backspacing out with keystrokes).    

3. The addition of the `x` to clear a selected value improves many things 
- It is now a bit more clear what is selected Vs what is a search term
- There is an obvious way to clear the value 
- We're forcing the user to acknowledge they've cleared the value because they begin searching which to me makes it more obvious what's happening

With these changes, I feel okay with the user picker. 👍 

@raghavaggarwal2308 I think we're good on the above ☝️ 

I see an unrelated issue where on 1st login the user does not get the icon. The webapp must be refreshed. This does not only happen on HA.   
image

I don't know if this will get solved when we cut a release or maybe it has already existed. This change must still be cherry-picked into the 2.x.x brach correct? If so, maybe I should just open a new issue if I see it occurring there?

raghavaggarwal2308:
@DHaussermann What do you exactly mean by the first login here? We tried re-deploying the plugin and connecting again but the issue is not reproducible on our local.

DHaussermann:
hi @raghavaggarwal2308 
When you create an account and login to Mattermost for the first time. It's possible this is a broader issue with App Bar but I don't think it's happening for the other icons. 

1. Install and enable the SN plugin
2. Get an invite link to a team (can use https://dkh-service-now.test.mattermost.cloud/signup_user_complete/?id=b71wyfc76bbidxg4pas8syocho)
3. Follow the link (in a new browser session if you're already logged in)
4. Create an account and login
**Observed**: The SN app bar icon is missing until you refresh.

DHaussermann:
@raghavaggarwal2308  one more note on this ☝️ 

I've been using `6.3.0` as a test server. But I'm not sure that's valid as I hit other unrelated webapp issues. 

That being said - I do see this on the last 6.3.x release as well `6.3.6`

image



Also confirmed another App Bar plugin Zoom does not seem affected.

raghavaggarwal2308:
@DHaussermann The icon issue should be fixed now. Can you please confirm?

DHaussermann (approved):
Tested and passed
Icon and all other know issues are now resolved.

LGTM!

DHaussermann:
@raghavaggarwal2308 Thanks for the effort on this. I have 2 related questions that do no block the PR.

1. Do you know what's happening with the App Bar? This plugin shows in the app bar even when it does not seem enabled for  other plugins (see below) I imagine this is intentional. Do you recall anything about this plugin always showing in the app bar?
image

2. Can you please clarify for me if this PR will now be cherry picked to a v`2.x.x` branch?

raghavaggarwal2308:
@DHaussermann 
1. Not sure what you are saying here. The icon is not visible when we disable the plugin or the app bar. Can you please send us the steps to reproduce the issue?
2. The issue #172 was to move the create incident dialog from v3.x.x to v2.x.x. So, once this PR is merged we will create a new v2.x.x release.

DHaussermann:
1. Yes I suppose you have answered my question as that is the issue. 
To clarify, when the app bar is disabled, the icon will not show in the channel header the way the other icons do. This must be by design and I simply don't recall. Can you find a technical reason why the icon does not show in the channel header? This was done intentionally correct?

2. Thanks. I understand 👍

DHaussermann:
@raghavaggarwal2308 sorry I missed the at mention for you ☝️

raghavaggarwal2308:
@DHaussermann The icon will be visible in the channel header when the app bar is disabled but the icon color will be changed. Please check the below screenshots. Let me know if you are unable to see the icon.
Screenshot from 2023-06-23 21-09-19
Screenshot from 2023-06-23 21-09-32

DHaussermann:
@raghavaggarwal2308 please ping mw once we have this merged and packaged into a release for release tesing.

raghavaggarwal2308:
@mickmister Can you please merge this PR if nothing is pending?

server/plugin/util.go
```diff
@@ -308,23 +308,33 @@ func decodeKey(key string) (string, error) {
...
+	if channel.Type == model.ChannelTypeDirect || channel.Type == model.ChannelTypeGroup {
+		return http.StatusBadRequest, fmt.Errorf(constants.ErrorInvalidChannelType)
 	}
 
 	// Check if a user is a part of the channel
 	if _, channelErr := p.API.GetChannelMember(channelID, userID); channelErr != nil {
 		p.API.LogDebug(constants.ErrorChannelPermissionsForUser, "Error", channelErr.Error())
-		return channelErr.StatusCode, "", fmt.Errorf(constants.ErrorInsufficientPermissions)
+		return channelErr.StatusCode, fmt.Errorf(constants.ErrorInsufficientPermissions)
 	}
```
mickmister: Same request here to use `HasPermissionToChannel`

webapp/src/containers/createIncident/subscribeToNewIncident.tsx
```diff
@@ -0,0 +1,100 @@
...
+    showChannelPanel,
+    setShowChannelPanel}: PropTypes) => {
+    // usePluginApi hook
+    const {getApiState} = usePluginApi();
+
+    const dispatch = useDispatch();
+
+    const getSubscriptionState = () => {
+        const {isLoading, isSuccess, isError, data, error: apiErr} = getApiState(Constants.pluginApiServiceConfigs.createSubscription.apiServiceName, subscriptionPayload);
+        return {isLoading, isSuccess, isError, data: data as RecordData, error: (apiErr as FetchBaseQueryError)?.data as APIError | undefined};
```
mickmister, line 52: Can we split this up into more statements here instead of putting everything in one deconstruction and one return statement? It's not clear what's happening with all of the `as`'s.

Can `getApiState` handle the casting instead?

webapp/src/containers/createIncident/subscribeToNewIncident.tsx
```diff
@@ -0,0 +1,100 @@
...
+                handleError(error);
+            }
+
+            if (isSuccess) {
+                setApiError(null);
+                dispatch(refetch());
+                setShowResultPanel(true);
+            }
+        }
+    }, [getSubscriptionState().isError, getSubscriptionState().isSuccess, getSubscriptionState().isLoading]);
```
mickmister, line 69: We could call `getSubscriptionState()` once instead of calling it 3 times here. Also, the `useEffect` callback should not be calling `getSubscriptionState` since that is "recreating" the dependencies for the callback.

Also `subscriptionPayload` needs to be added to the dependency array

```ts
const {isLoading, isError, isSuccess, error} = getSubscriptionState();

useEffect(() => {
    if (subscriptionPayload) {
        setShowModalLoader(isLoading);
        if (isError && error) {
            handleError(error);
        }

        if (isSuccess) {
            setApiError(null);
            dispatch(refetch());
            setShowResultPanel(true);
        }
    }
}, [subscriptionPayload, isError, isSuccess, isLoading]);
```

server/plugin/api_test.go
```diff
@@ -1842,3 +1843,138 @@ func TestCheckSubscriptionsConfigured(t *testing.T) {
...
+			},
+			SetupPlugin:        func(p *Plugin) {},
+			SetupClient:        func(client *mock_plugin.Client) {},
+			ExpectedStatusCode: http.StatusBadRequest,
+		},
+		"does not have permission to access the channel": {
+			RequestBody: testutils.GetCreateIncidentPayload(),
+			SetupAPI:    func(api *plugintest.API) {},
+			SetupPlugin: func(p *Plugin) {
+				monkey.PatchInstanceMethod(reflect.TypeOf(p), "HasChannelPermissions", func(_ *Plugin, _, _ string) (int, error) {
```
mickmister, line 1904: Ideally we mock the plugin API using `api.On`, instead of mocking the plugin's code, in order to have a more realistic test

webapp/src/app.tsx
```diff
@@ -1,16 +1,33 @@
...
         makeApiRequest(Constants.pluginApiServiceConfigs.getConfig.apiServiceName);
+        makeApiRequest(Constants.pluginApiServiceConfigs.getConnectedUser.apiServiceName);
     }, []);
 
+    useEffect(() => {
+        const {data, isLoading} = getConnectedUserState();
+        if (!isLoading && data) {
+            dispatch(setConnected(data.connected));
+        }
+    }, [getConnectedUserState().isLoading, getConnectedUserState().data]);
```
mickmister, line 29: Same request to de-duplicate calls to `getConnectedUserState`, and avoid recomputing state inside of `useEffect` callback

mickmister (commented):
I gave the PR another review. In general looks good :+1: Just a few comments for discussion

webapp/src/containers/createIncident/createIncidentMenu.tsx
raghavaggarwal2308: @mickmister It's very uncommon for the SiteURL and the browser's URL to be different. We have seen in the past that in a few plugins, the functionality breaks if the URLs are different. So, most of the servers keep these URLs the same. Is there really a need for this change? Because it will add unnecessary complexity to the code.

webapp/src/utils/index.tsx
```diff
@@ -15,8 +16,7 @@ const getBaseUrls = (): {
     mattermostApiBaseUrl: string;
     publicFilesUrl: string;
 } => {
-    const url = new URL(window.location.href);
-    const baseUrl = `${url.protocol}//${url.host}`;
+    const baseUrl = Cookies.get(Constants.SiteUrl);
```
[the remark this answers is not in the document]
raghavaggarwal2308: @mickmister In context of [this comment](https://github.com/mattermost/mattermost-plugin-servicenow/pull/182#discussion_r1256077015):
There won't be any proper way to fetch the site URL from redux every time and send it in the `getBaseUrl`, we need to use some hacks to achieve this and it will get very complicated to handle all edge cases. Also, we cannot defer the `createApi` call.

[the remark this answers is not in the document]
ayusht2810: @mickmister we have found the solution for the above issue and currently working on it.

raghavaggarwal2308:
@mickmister FIxed the review comments mentioned by you

webapp/src/containers/createIncident/createIncidentMenu.tsx
```diff
@@ -21,6 +21,7 @@ const CreateIncidentPostMenuAction = ({postId}: PropTypes) => {
     const {pluginState} = usePluginApi();
     const dispatch = useDispatch();
     const post = useSelector((state: GlobalState) => getPost(state, postId));
+    const {SiteURL} = useSelector((state: GlobalState) => state.entities.general.config);
```
mickmister, line 24: Can we put this in a separate function to avoid duplicating this logic?

```suggestion
    const siteURL = useSelector(getSiteURL);
```

webapp/src/utils/index.tsx
```diff
@@ -3,23 +3,21 @@
...
-import Cookies from 'js-cookie';
 
 import {Button} from '@brightscout/mattermost-ui-library';
 
 import Constants, {SubscriptionType, RecordType, KnowledgeRecordDataLabelConfigKey, RecordDataLabelConfigKey, CONNECT_ACCOUNT_LINK} from 'src/plugin_constants';
 
 import {id as pluginId} from '../manifest';
 
-const getBaseUrls = (): {
+const getBaseUrls = (mmSiteUrl?: string): {
```
mickmister, line 13: Easiest way to default to empty string. If we don't default to empty string, `undefined` will show up in the outout

```suggestion
const getBaseUrls = (mmSiteUrl = ''): {
```

webapp/src/services/index.ts
```diff
@@ -1,13 +1,40 @@
+// eslint-disable-next-line import/no-unresolved
+import {BaseQueryApi} from '@reduxjs/toolkit/dist/query/baseQueryTypes';
```
mickmister, line 2: Is this type still useful if the import can't be resolved?

webapp/src/index.tsx
```diff
@@ -47,10 +46,7 @@ export default class Plugin {
         registry.registerChannelHeaderButtonAction(<ServiceNowIcon className='servicenow-icon'/>, () => store.dispatch(toggleRHSPlugin), null, Constants.ChannelHeaderTooltipText);
         registry.registerAdminConsoleCustomSetting('ServiceNowUpdateSetDownload', DownloadButton);
 
-        // Store the Mattermost site URL in cookies
-        Cookies.set(Constants.SiteUrl, store.getState().entities.general.config.SiteURL as string);
-
-        const iconUrl = `${Utils.getBaseUrls().publicFilesUrl}${Constants.SERVICENOW_ICON_URL}`;
+        const iconUrl = `${Utils.getBaseUrls(store.getState().entities.general.config.SiteURL).publicFilesUrl}${Constants.SERVICENOW_ICON_URL}`;
```
mickmister, line 49: Probably good to avoid accessing `state.entities` etc. in this sort of context too.

```suggestion
        const iconUrl = `${Utils.getBaseUrls(getSiteURL(store.getState()).publicFilesUrl}${Constants.SERVICENOW_ICON_URL}`;
```

mickmister (approved):
LGTM, just some thoughts for improvement

webapp/src/services/index.ts
raghavaggarwal2308: @mickmister  Yes, it's useful, we are using it as the type of parameter in `handleBaseQuery` function.

raghavaggarwal2308:
@mickmister Fixed the review comments mentioned by you

webapp/src/services/index.ts
mickmister: But if it can't be resolved, how does it work? Is it the case that your editor can resolve it, but for whatever reason resolver being used by eslint cannot find it?

mickmister:
> @raghavaggarwal2308 please ping mw once we have this merged and packaged into a release for release testing.

Merging due to @DHaussermann's approval above

webapp/src/services/index.ts
raghavaggarwal2308: @mickmister Yes, it is being resolved by the editor but not by eslint.

mickmister: I wonder why that's the case. I would prefer we didn't have to do this, but it's pretty negligible in this case

PR #190, opened 2024-01-02
[MM-64] Update plugin with respect to phase 1 upgrades

ayusht2810:
#### Summary
Updated plugin with respect to phase 1 upgrades

- Updated `build` folder to sync with starter template
- Updated package in go modules to use https://pkg.go.dev/github.com/mattermost/mattermost/server/public
- Updated Makefile
- Updated version in `nvmrc` file and update `package-lock.json` file.
- Update test cases base setup
- Updated developer docs link in the readme file 

#### Ticket Link


webapp/src/types/mattermost-webapp/index.d.ts
```diff
@@ -1,6 +1,4 @@
-import React from 'react';
-
-export interface PluginRegistry {
+interface PluginRegistry {
     registerReducer(reducer)
     registerPostTypeComponent(typeName: string, component: React.ElementType)
```
mickmister, line 3: Really not a fan of this `d.ts` stuff. Much prefer the explicit import/export style. No need to change this here though

webapp/src/websocket/index.ts
```diff
@@ -11,10 +11,7 @@ import {refetch} from 'src/reducers/refetchState';
 export function handleConnect(store: Store<GlobalState, Action<Record<string, unknown>>>, rhsComponentId: string) {
     return (_: WebsocketEventParams) => {
         store.dispatch(setConnected(true) as Action);
-        const {rhsState, pluggableId} = (store.getState() as GlobalState).views.rhs;
-        if (rhsState === 'plugin' && pluggableId === rhsComponentId) {
-            store.dispatch(refetch() as Action);
-        }
+        store.dispatch(refetch() as Action);
     };
```
mickmister, line 15: This is a change in logic. Is this a safe change?

webapp/src/websocket/index.ts
```diff
@@ -51,10 +48,7 @@ export function handleOpenEditSubscriptionModal(store: Store<GlobalState, Action
 
 export function handleSubscriptionDeleted(store: Store<GlobalState, Action<Record<string, unknown>>>, rhsComponentId: string) {
     return (_: WebsocketEventParams) => {
-        const {rhsState, pluggableId} = (store.getState() as GlobalState).views.rhs;
-        if (rhsState === 'plugin' && pluggableId === rhsComponentId) {
-            store.dispatch(refetch() as Action);
-        }
+        store.dispatch(refetch() as Action);
     };
```
mickmister, line 52: Same here

webapp/src/websocket/index.ts
ayusht2810: This was needed as we don't have any `views` present in the `GlobalState` [type](https://github.com/mattermost/mattermost-redux/blob/42a477837cdb1461b25dc30695a3ac3d951accca/src/types/store.ts#L25).

Also, this logic was used to help in dispatching the `refetch` action only when the plugin's RHS is enabled.

mickmister: Can we extend the `GlobalState` type to include the views? It seems like this is an unnecessary performance hit on a websocket event

Kshitij-Katiyar: @mickmister From my understanding I don't think we can extend this as we don't have the views present anywhere in the Global state of the mattermost redux.

mickmister: But we can change the type to have the views right? Or are you saying that the old code is broken and the values simply don't exist in the redux store? I suppose those are subject to change at any time too so probably good not to rely on them. cc @hmhealey in case you have any thoughts about plugins accessing the `views` state of redux, and also the likely frowned-upon modifying of the type to expose that slice to the type system

mickmister: I still think the logic is needed here right?

ayusht2810: @mickmister Yes, the logic is needed here to reduce the number of `refetches`. But as stated in the https://github.com/mattermost/mattermost-plugin-servicenow/pull/190#discussion_r1490979162 we don't have views present in the redux state. The logic will still work fine, but there is an increase in some unecessary `refetches`.

mickmister: @ayusht2810 How about we try this:

- We assume the views *might* be there (make a special type for the `GlobalStore` used only in this spot, that has this information "optional", meaning it may be `undefined`)
- Check if the `rhsState` is available in the store
  - If available, do the logic that was here before
  - If not, do the logic here that does not access `rhsState`

hmhealey: My feedback is a bit late here, but I think keeping the previous version of this code is probably fine. @mickmister is right that I'd recommend not accessing `state.views` from a plugin because we consider that to be web app state which we're more likely to change than `state.entities`, but we don't have any plans to do that any time soon.

Ideally, the plugin's RHS component would register its own WS handler instead of having that be registered in the plugin's `initialize` method since it's presumably the only thing that cares about the data loaded by `refetch`. That could be done using `window.ProductApi.useWebSocketClient`, but I don't know if this plugin is one that we'd want to use the `ProductApi` for since that API is a bit more fragile than the regular plugin one

Kshitij-Katiyar: updated the condition to check for `rhsState` and `pluggableId` when refetching.
@mickmister @hmhealey  what is your opinion on this?

mickmister: @Kshitij-Katiyar I don't see this change in the PR

Kshitij-Katiyar: Updated the condition for refetch based upon the `rhsState` and `pluggableID`.
@mickmister @hmhealey What's your opinion on this?

**commit Link:** https://github.com/mattermost/mattermost-plugin-servicenow/pull/190/commits/6e075462c965f782625bfbfc0812bef7bededa74

webapp/src/types/common/globalState.ts
```diff
@@ -0,0 +1,5 @@
+import {GlobalState as OriginalGlobalState} from 'mattermost-redux/types/store';
+
+export interface GlobalState extends OriginalGlobalState {
+    views?: any;
```
mickmister, line 4: I'm thinking this will be more helpful with typescript

```suggestion
    views?: {
        rhs?: {
            rhsState: string;
            pluggableId: string;
        };
    };
```

Kshitij-Katiyar: @mickmister Updated the PR with these changes

webapp/src/websocket/index.ts
```diff
@@ -11,8 +11,10 @@ import {refetch} from 'src/reducers/refetchState';
 export function handleConnect(store: Store<GlobalState, Action<Record<string, unknown>>>, rhsComponentId: string) {
     return (_: WebsocketEventParams) => {
         store.dispatch(setConnected(true) as Action);
-        const {rhsState, pluggableId} = (store.getState() as GlobalState).views.rhs;
-        if (rhsState === 'plugin' && pluggableId === rhsComponentId) {
+        const globalState = (store.getState() as GlobalState);
+        if (globalState.views && globalState.views?.rhs?.rhsState === 'plugin' && globalState.views?.rhs?.pluggableId === rhsComponentId) {
+            store.dispatch(refetch() as Action);
+        } else {
             store.dispatch(refetch() as Action);
```
mickmister, line 18: I think we should avoid fetching in the `else` right? Same with the other block in this PR

mickmister:
```suggestion
        if (globalState.views?.rhs && 'rhsState' in globalState.views?.rhs) {
            if (globalState.views?.rhs?.rhsState === 'plugin' && globalState.views?.rhs?.pluggableId === rhsComponentId) {
                store.dispatch(refetch() as Action);
            }
        } else {
            store.dispatch(refetch() as Action);
```

ayusht2810:
@mickmister Fixed the review comments. Please re-review.

webapp/src/websocket/index.ts
```diff
@@ -11,8 +11,12 @@ import {refetch} from 'src/reducers/refetchState';
...
-        const {rhsState, pluggableId} = (store.getState() as GlobalState).views.rhs;
-        if (rhsState === 'plugin' && pluggableId === rhsComponentId) {
+        const globalState = (store.getState() as GlobalState);
+        if (globalState.views?.rhs?.rhsState === 'plugin') {
+            if (globalState.views.rhs.pluggableId === rhsComponentId) {
+                store.dispatch(refetch() as Action);
+            }
+        } else {
             store.dispatch(refetch() as Action);
         }
```
mickmister, line 21: We should make it so this  `else` essentially "never" happens. Meaning it should only happen if the redux store has changed its structure from what we currently assume. The first if should just be "does the redux structure match our expectations", and not make any assertions about the state of the store, as described in this comment https://github.com/mattermost/mattermost-plugin-servicenow/pull/190#discussion_r1645408180

ayusht2810: @mickmister Not sure whether I got your point or not. If the condition `if (globalState.views?.rhs?.rhsState === 'plugin')` is false, how will we refetch if we remove the else condition?

mickmister: @ayusht2810 As far as I understand, we only want to refetch if the plugin's RHS is currently showing. The `else` was added here in the case of "the redux store's structure itself changes in future iterations of the webapp codebase". That's a corner case that will essentially "never" be used, because the redux store structure will likely not change.

I'm not saying we should remove the `else` condition. I'm saying that the branch that runs that _will_/_should_ not happen unless the redux store structure changes.

This is how I think it should be implemented (verbosely):

```ts
const doesReduxStoreStructureMatchOurAssumptions = Boolean(globalState.views?.rhs && 'rhsState' in globalState.views?.rhs);
if (doesReduxStoreStructureMatchOurAssumptions) {
    if (globalState.views?.rhs?.rhsState === 'plugin' && globalState.views?.rhs?.pluggableId === rhsComponentId) {
        store.dispatch(refetch() as Action);
    }
} else {
    store.dispatch(refetch() as Action);
}
```

Personally I think we can get rid of the `else` altogether, but the undocumented redux store structure could in theory change in the future, so it's not a bad idea to have code that defends against it. I'm 0/5 on if we need it, but we should only refetch the data when necessary given what the code has to work with. @ayusht2810 Does this all make sense?

raghavaggarwal2308: @mickmister Just for clarification:
- The `if` block present here for handling the older redux structure (i.e. for order MM versions), where we had access to `globalState.views`.
- The `else` block is for handling the current redux structure with no access to `globalState.views`(i.e. latest MM version). So, if we remove the else block the RHS will not update until the user re-open it.

So, I do't think we should remove the else block as it will break the functionality on latest servers.

raghavaggarwal2308: cc: @wiggin77

mickmister: > The else block is for handling the current redux structure with no access to globalState.views(i.e. latest MM version). So, if we remove the else block the RHS will not update until the user re-open it.

@raghavaggarwal2308 Are we sure about this? I thought it was just that the exported `GlobalStore` type doesn't expose it. But I could have that incorrect

raghavaggarwal2308: @mickmister Yes we are sure, the field does not exist in redux in the latest version.

wiggin77:
@ayusht2810 has QA done any testing on this PR?  It is quite large to merge without some testing.

raghavaggarwal2308:
@wiggin77 We generally assign a PR for QA once the PR is dev approved. If the changes looks good to you we can assign it to @AayushChaudhary0001

hmhealey:
Dismissing my request for review since I think I was just added to this PR to answer a question, and there's enough other devs assigned at the moment. I've been a bit swamped with big PRs to review recently, but feel free to re-add me if needed

AayushChaudhary0001 (approved):
I have tested this PR and all the basic functionalities are working fine, LGTM. Approved.

raghavaggarwal2308:
@wiggin77 Can you give this OR a review? It is requred for the secret config PR https://github.com/mattermost/mattermost-plugin-servicenow/pull/201

wiggin77 (approved):

raghavaggarwal2308 (approved):

raghavaggarwal2308:
@wiggin77 The CI is not running on this PR. I think this might be due to [an outdated branch protection rule](https://github.com/orgs/community/discussions/26698#discussioncomment-10551267). Can you check the same

Issue #163, opened 2023-01-11
Add required Mattermost Server Version to documentation


asatkinson:
Currently this is 5.37

PR #191, opened 2024-03-15
[GH-163]: Added the minimum supported mm version in the docs

Kshitij-Katiyar:
#### Summary
Updated the read me to add the minimum supported mattermost version in the docs which is v6.3.0.
Issue: https://github.com/mattermost/mattermost-plugin-servicenow/issues/163https://github.com/mattermost/mattermost-plugin-servicenow/issues/163

### DIFF
--- a/docs/developer_docs.md
+++ b/docs/developer_docs.md
@@ -139,6 +139,8 @@ Make sure you have the following components installed:
 
 - Make
 
+- The minimum supported version for mattermost is v6.3.0.
+
 ### Building the plugin
 
 Run the following command in the plugin repo to prepare a compiled, distributable plugin zip:


docs/developer_docs.md
```diff
@@ -139,6 +139,8 @@ Make sure you have the following components installed:
 
 - Make
 
+- The minimum supported version for mattermost is v6.3.0.
+
```
mickmister, line 143: I'm concerned about having this in the readme, since it will easily be made outdated if we change the `min_server_version`, and we will need to remember to change this. Is this coming up for a specific reason?

Kshitij-Katiyar: @mickmister It was a request of a user https://github.com/mattermost/mattermost-plugin-servicenow/issues/163. I can close this PR as well issue if this is not needed.

mickmister: Yeah I think we should close the PR. There's little value imo for having this in the README. It ends up being outdated information that we delete instead of update

mickmister: Also FYI that issue was created by a previous employee of Mattermost

PR #202, opened 2024-11-29
Remove schedule from CI

toninis:
#### Summary

Remove schedule from CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,5 @@
 name: ci
 on:
-  schedule:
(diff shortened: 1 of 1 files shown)


BenCookie95 (commented):
Why are we removing this from this plugin? A lot of plugins I checked have this schedule set up

toninis:
> Why are we removing this from this plugin? A lot of plugins I checked have this schedule set up

We do not need it . Plus if there is no activity for 60 days on the repo the CI is locked .

BenCookie95 (approved):
Sounds good, thanks for the explanation!

